PR #2, opened 2021-05-08
Fixes gradient, Jacobian, Hessian, and vjp tests

frankschae:
Current test status:
```julia
Test Summary:              | Pass  Fail  Error  Total
AbstractDifferentiation.jl |  103     1      1    105
  FiniteDifferences        |  103     1      1    105
    Derivative             |   12                  12
    Gradient               |   18                  18
    Jacobian               |   18                  18
    Hessian                |   30                  30
    jvp                    |    7     1      1      9
    j′vp                   |   18                  18
```
There are basically two fixes:
- FiniteDifferences mutates the input `x` to compute the gradients, so the value was computed wrongly. 
- when the identity was an `AbstractMatrix` it didn't loop over its columns but vectorized the entire matrix which lead to a dimension error. 

It remains to fix the jvp case.
```julia
test_fdm_jvp(fdm_backend1) # passes
test_fdm_jvp(fdm_backend2) # fails
test_fdm_jvp(fdm_backend3) # fails
```

For example, in `fdm_backend2`
```julia
v = (rand(length(xvec)), rand(length(yvec)))
pf1 = AD.pushforward_function(fdm_backend2, fjac, xvec, yvec)(v) 
```
results in a *single* output vector instead of a *tuple of two*  output vectors.

### DIFF
--- a/src/AbstractDifferentiation.jl
+++ b/src/AbstractDifferentiation.jl
@@ -305,15 +305,18 @@ struct LazyJacobian{B, F, X}
 end
 
 function Base.:*(d::LazyJacobian, ys)
+    if !(ys isa Tuple)
+        ys = (ys, )
+    end
     if d.xs isa Tuple
-        jvp = pushforward_function(d.backend, d.f, d.xs...)(ys)
-        if jvp isa Tuple
-            return vec.(jvp)
-        else
-            return vec(jvp)
-        end
+        vjp = pushforward_function(d.backend, d.f, d.xs...)(ys)
     else
-        return vec.(pushforward_function(d.backend, d.f, d.xs)(ys))
+        vjp = pushforward_function(d.backend, d.f, d.xs)(ys)
+    end
+    if vjp isa Tuple
+        return vjp
+    else
+        return (vjp,)
     end
 end
 
@@ -346,19 +349,6 @@ function Base.:*(ys::Number, d::LazyJacobian)
     end
 end
 
-function Base.:*(d::LazyJacobian, ys::AbstractArray)
-    if d.xs isa Tuple
-        return vec.(pushforward_function(d.backend, d.f, d.xs...)((ys,)))
-    else
-        vjp = (pushforward_function(d.backend, d.f, d.xs)((ys,)))
-        if vjp isa Tuple
-            return vec.(vjp)
-        else
-            return (vec(vjp),)
-        end
-    end
-end
-
 
 struct LazyHessian{B, F, X}
     backend::B
@@ -367,60 +357,22 @@ struct LazyHessian{B, F, X}
 end
 
 function Base.:*(d::LazyHessian, ys)
+    if !(ys isa Tuple)
+        ys = (ys, )
+    end
+
     if d.xs isa Tuple
         return pushforward_function(
             secondlowest(d.backend),
-                (xs...,) -> gradient(lowest(d.backend), d.f, xs...), d.xs...,)(ys)
+            (xs...,) -> gradient(lowest(d.backend), d.f, xs...), d.xs...,)(ys)
     else
         return pushforward_function(
             secondlowest(d.backend),
-                (xs,) -> gradient(lowest(d.backend), d.f, xs),d.xs,)(ys)
+            (xs,) -> gradient(lowest(d.backend), d.f, xs),d.xs,)(ys)
     end
-
 end
 
 function Base.:*(ys, d::LazyHessian)
-    if ys isa Tuple
-        ya = adjoint.(ys)
-    else
-        ya = adjoint(ys)
-    end
-    return pullback_function(
-        secondlowest(d.backend),
-        (xs...,) -> gradient(lowest(d.backend), d.f, xs...),
-        d.xs...,
-    )(ya)
-end
-
-function Base.:*(d::LazyHessian, ys::Number)
-    if d.xs isa Tuple
-        return hessian(d.backend, d.f, d.xs...).*ys
-    else
-        return hessian(d.backend, d.f, d.xs).*ys
-    end
-end
-
-function Base.:*(ys::Number, d::LazyHessian)
-    if d.xs isa Tuple
-        return ys.*hessian(d.backend, d.f, d.xs...)
-    else
-        return ys.*hessian(d.backend, d.f, d.xs)
-    end
-end
-
-function Base.:*(d::LazyHessian, ys::AbstractArray)
-    if d.xs isa Tuple
-        return pushforward_function(
-            secondlowest(d.backend),
-                (xs...,) -> gradient(lowest(d.backend), d.f, xs...), d.xs...,)((ys,))
-    else
-        return pushforward_function(
-            secondlowest(d.backend),
-                (xs,) -> gradient(lowest(d.backend), d.f, xs),d.xs,)((ys,))
-    end
-end
-
-function Base.:*(ys::AbstractArray, d::LazyHessian)
     if ys isa Tuple
         ya = adjoint.(ys)
     else
@@ -440,6 +392,23 @@ function Base.:*(ys::AbstractArray, d::LazyHessian)
             )(ya)
     end
 end
+
+function Base.:*(d::LazyHessian, ys::Number)
+    if d.xs isa Tuple
+        return hessian(d.backend, d.f, d.xs...).*ys
+    else
+        return hessian(d.backend, d.f, d.xs).*ys
+    end
+end
+
+function Base.:*(ys::Number, d::LazyHessian)
+    if d.xs isa Tuple
+        return ys.*hessian(d.backend, d.f, d.xs...)
+    else
+        return ys.*hessian(d.backend, d.f, d.xs)
+    end
+end
+
 
 function lazyderivative(ab::AbstractBackend, f, xs::Number...)
     return LazyDerivative(ab, f, xs)
@@ -511,7 +480,9 @@ function define_pushforward_function_and_friends(fdef)
                     end
                 end
             elseif eltype(identity_like) <: AbstractMatrix
+                # needed for the computation of the Hessian
                 ret = hcat.(mapslices(identity_like[1], dims=1) do cols
+                    # cols loop over basis states   
                     pf = pff((cols,))
                     if typeof(pf) <: AbstractVector
                         return (pf, )
--- a/src/AbstractDifferentiation.jl
+++ b/src/AbstractDifferentiation.jl
@@ -31,8 +31,14 @@ primalvalue(x::Tuple) = map(primalvalue, x)
 primalvalue(x) = x
 
 function derivative(ab::AbstractBackend, f, xs::Number...)
-    return getindex.(jacobian(lowest(ab), f, xs...), 1)
+    der = getindex.(jacobian(lowest(ab), f, xs...), 1)
+    if der isa Tuple
+        return der
+    else
+        return (der,)
+    end
 end
+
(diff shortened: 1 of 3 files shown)


src/AbstractDifferentiation.jl
```diff
@@ -54,6 +54,10 @@ end
 function value_and_jacobian(ab::AbstractBackend, f, xs...)
     local value
     primalcalled = false
+    if ab isa AbstractFiniteDifference
```
mohamed82008, line 57:
```suggestion
    if lowest(ab) isa AbstractFiniteDifference
```

src/AbstractDifferentiation.jl
```diff
@@ -54,6 +54,10 @@ end
...
+        value = primalvalue(ab, nothing, f, xs)
```
mohamed82008, line 58:
```suggestion
        value = primalvalue(lowest(ab), nothing, f, xs)
```

src/AbstractDifferentiation.jl
```diff
@@ -224,63 +247,200 @@ struct LazyDerivative{B, F, X}
...
-    return derivative(d.ab, d.f, d.xs...) * y
+    return derivative(d.backend, d.f, d.xs...) * y
 end
+
 function Base.:*(y, d::LazyDerivative)
-    return y * derivative(d.ab, d.f, d.xs...)
+    return y * derivative(d.backend, d.f, d.xs...)
+end
+
+function Base.:*(d::LazyDerivative, y::Union{Number,Tuple})
```
mohamed82008, line 259: I don't think we should support multiplication if the derivative returns a tuple. This part of the code can be simplified.

src/AbstractDifferentiation.jl
```diff
@@ -224,63 +247,200 @@ struct LazyDerivative{B, F, X}
...
     f::F
     xs::X
 end
-Base.:*(d::LazyGradient, y) = gradient(d.ab, d.f, d.xs...) * y
-Base.:*(y, d::LazyGradient) = y * gradient(d.ab, d.f, d.xs...)
+Base.:*(d::LazyGradient, y) = gradient(d.backend, d.f, d.xs...) * y
+Base.:*(y, d::LazyGradient) = y * gradient(d.backend, d.f, d.xs...)
+
+function Base.:*(d::LazyGradient, y::Union{Number,Tuple})
+    if d.xs isa Tuple
```
mohamed82008, line 285: Why is `xs` sometimes a tuple and sometimes not?

frankschae: I wanted to support:
 `AD.LazyJacobian(fdm_backend, x->fjac(x, yvec), xvec)`
where `xs` is then only a vector `xvec` and not (xvec, ).

src/AbstractDifferentiation.jl
frankschae: Don't we want to support functions like `fder(x, y) = exp(y) * x + y * log(x)` for the LazyOperators as well?

mohamed82008: Hmm I guess we could.

src/AbstractDifferentiation.jl
mohamed82008: I see.

src/AbstractDifferentiation.jl
```diff
@@ -224,63 +247,200 @@ struct LazyDerivative{B, F, X}
...
     f::F
     xs::X
 end
+
 function Base.:*(d::LazyJacobian, ys)
-    return pushforward_function(d.ab, d.f, d.xs...)(ys)
+    if d.xs isa Tuple
+        jvp = pushforward_function(d.backend, d.f, d.xs...)(ys)
+        if jvp isa Tuple
+            return vec.(jvp)
```
mohamed82008, line 311: why are we calling `vec`?

src/AbstractDifferentiation.jl
```diff
@@ -224,63 +247,200 @@ struct LazyDerivative{B, F, X}
...
+    else
+        return jacobian(d.backend, d.f, d.xs) .* ys
+    end
+end
+
+function Base.:*(d::LazyJacobian, ys::AbstractArray)
+    if d.xs isa Tuple
+        return vec.(pushforward_function(d.backend, d.f, d.xs...)((ys,)))
+    else
+        vjp = (pushforward_function(d.backend, d.f, d.xs)((ys,)))
```
mohamed82008, line 353: unnecessary outer parenthesis

src/AbstractDifferentiation.jl
```diff
@@ -224,63 +247,200 @@ struct LazyDerivative{B, F, X}
...
 function Base.:*(d::LazyHessian, ys)
-    return pushforward_function(
-        secondlowest(d.ab),
-        (xs...,) -> gradient(lowest(d.ab), d.f, xs...),
-        d.xs...,
-    )(ys)
+    if d.xs isa Tuple
+        return pushforward_function(
+            secondlowest(d.backend),
+                (xs...,) -> gradient(lowest(d.backend), d.f, xs...), d.xs...,)(ys)
```
mohamed82008, line 373: nitpick: inconsistent line formatting

src/AbstractDifferentiation.jl
```diff
@@ -224,63 +247,200 @@ struct LazyDerivative{B, F, X}
...
 
+function Base.:*(ys::Number, d::LazyJacobian)
+    if d.xs isa Tuple
+        return jacobian(d.backend, d.f, d.xs...) .* ys
+    else
+        return jacobian(d.backend, d.f, d.xs) .* ys
+    end
+end
+
+function Base.:*(d::LazyJacobian, ys::AbstractArray)
```
mohamed82008, line 349: I don't think this does the right thing if `ys` is a matrix.

src/AbstractDifferentiation.jl
```diff
@@ -224,63 +247,200 @@ struct LazyDerivative{B, F, X}
...
+
+function Base.:*(ys::Number, d::LazyHessian)
+    if d.xs isa Tuple
+        return ys.*hessian(d.backend, d.f, d.xs...)
+    else
+        return ys.*hessian(d.backend, d.f, d.xs)
+    end
+end
+
+function Base.:*(d::LazyHessian, ys::AbstractArray)
```
mohamed82008, line 411: if `ys` is a matrix, this doesn't seem right.

mohamed82008: how is this function different from the general fallback?

src/AbstractDifferentiation.jl
```diff
@@ -224,63 +247,200 @@ struct LazyDerivative{B, F, X}
...
+            secondlowest(d.backend),
+                (xs...,) -> gradient(lowest(d.backend), d.f, xs...), d.xs...,)((ys,))
+    else
+        return pushforward_function(
+            secondlowest(d.backend),
+                (xs,) -> gradient(lowest(d.backend), d.f, xs),d.xs,)((ys,))
+    end
+end
+
+function Base.:*(ys::AbstractArray, d::LazyHessian)
```
mohamed82008, line 423: can we combine this with the general fallback?

src/AbstractDifferentiation.jl
```diff
@@ -350,6 +510,19 @@ function define_pushforward_function_and_friends(fdef)
                         pff(cols)
                     end
                 end
+            elseif eltype(identity_like) <: AbstractMatrix
```
mohamed82008, line 513: Could you please comment this part? It's not clear to me what's happening here.

test/runtests.jl
```diff
@@ -99,19 +139,42 @@ function test_fdm_hessians(fdm_backend)
...
+    fhess2 = x-> dfgraddx(x, yvec)
+    hess5 = AD.jacobian(fdm_backend, fhess2, xvec)
+    @test minimum(isapprox.(hess5, hess1, atol=1e-10))
 end
 
 function test_fdm_jvp(fdm_backend)
     v = (rand(length(xvec)), rand(length(yvec)))
-    pf1 = AD.pushforward_function(fdm_backend, fjac, xvec, yvec)(v)
+
+    if fdm_backend isa FDMBackend2 # augmented version of v
```
mohamed82008, line 153: Could you explain this part?

mohamed82008:
Looks good overall. Just left a few comments because I couldn't immediately understand what's happening in a couple of places.

src/AbstractDifferentiation.jl
frankschae: Should we throw an error in that case or fix that manually by putting a `vec`?  I now combined  these `AbstractArray` specializations with the general fallback. They were just differing by stuff like `(y,)`.

src/AbstractDifferentiation.jl
frankschae: good idea! See my comment above as well -- When I started I thought it makes sense to dispatch them all separately because of these tiny differences between Arrays, Tuples, and Numbers as input but it really leads to much more code and gets less readable..

src/AbstractDifferentiation.jl
frankschae: nice catch -- was unnecessary here. I think I initially added it because of the `adjoint` with the pullback.

test/runtests.jl
frankschae: This is related to the comment above: https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/2#issuecomment-839138438 . 

The resulting error traces back to the following: When we use: `v = (rand(length(xvec)), rand(length(yvec)))` with our `FDMBackend2` as the vector that we'd like to pushforward through a  function `fvec(x,y)` that takes two inputs `x` and `y`, we get out an output that corresponds to the directional derivative with "`v[1]` as the direction" along x *and* "`v[2]` as the direction" along y. The other backends however interprete the `v` as change `x` along the direction `v[1]` while keeping `y` fix and  change `y` in the direction `v[2]` while keeping `x` fix.

src/AbstractDifferentiation.jl
frankschae: This was the fix I added for the computation of the Hessian. `identity_like` and `cols` look as follows in that case:
```julia
identity_like = ([1.0 0.0 0.0 0.0 0.0; 0.0 1.0 0.0 0.0 0.0; 0.0 0.0 1.0 0.0 0.0; 0.0 0.0 0.0 1.0 0.0; 0.0 0.0 0.0 0.0 1.0],)
identity_like[1] = [1.0 0.0 0.0 0.0 0.0; 0.0 1.0 0.0 0.0 0.0; 0.0 0.0 1.0 0.0 0.0; 0.0 0.0 0.0 1.0 0.0; 0.0 0.0 0.0 0.0 1.0]
cols = [1.0, 0.0, 0.0, 0.0, 0.0]
cols = [0.0, 1.0, 0.0, 0.0, 0.0]
cols = [0.0, 0.0, 1.0, 0.0, 0.0]
cols = [0.0, 0.0, 0.0, 1.0, 0.0]
cols = [0.0, 0.0, 0.0, 0.0, 1.0]
```
so it was mainly to fix the input/output for the additional `pushforward` that is used. I'll need to check in a bit more detail if one can simplify that function a bit more. IIRC without that function I got dimension errors in the jvp of FiniteDifferences.jl because it would have pushed forward matrices like `identity_like[1]` instead of the columns.

frankschae: but I think you are right that this needs some additional care.. I think the current version will break down once we'd like to compute a tuple of Hessians.

PR #5, opened 2021-08-29
Document API

mohamed82008:


mohamed82008:
I will add examples when the AD packages support AbstractDifferentiation.

README.md
```diff
@@ -1 +1,66 @@
...
+
+Julia has more (automatic) differentiation packages than you can count on 2 hands. Different packages have different user interfaces. Therefore, having a backend-agnostic interface to request the function value and its gradient for example is necessary to avoid a combinatorial explosion of code when trying to support every differentiation package in Julia in every algorithm package requiring gradients. For higher order derivatives, the situation is even more dire since you can combine any 2 differentiation backends together to create a new higher-order backend.
+
+## Loading `AbstractDifferentiation`
+
+To load `AbstractDifferentiation`, use:
+```julia
+using AbstractDifferentiation
+```
+`AbstractDifferentiation` exports a single name `AD` which is just an alias for the `AbstractDifferentiation` module itself. You can use this to access names inside `AbstractDifferentiation` using `AD.<>` instead of typing the long name `AbstractDifferentiation`.
```
odow, line 15: What's the point of this? Can't someone just do
```julia
import AbstractDifferentiation as AD
```
or 
```Julia
using AbstractDifferentiation
const AD = AbstractDifferentiation
```

README.md
```diff
@@ -1 +1,66 @@
...
+- `value_and_pb_f = AD.value_and_pullback_function(ab::AD.AbstractBackend, f, xs...)`: returns a function `value_and_pb_f` which accepts the co-tangent `vs` as input which is a tuple of length equal to the number of outputs of `f`. If `f` has a single output, `value_and_pb_f` can accept a single input instead of a 1-tuple. `value_and_pb_f` returns a 2-tuple, namely the value `f(xs...)` and output of the pullback operator.
+
+### Lazy operators
+
+You can also get a struct for the lazy derivative/gradient/Jacobian/Hessian of a function. You can then use the `*` operator to apply the lazy operator on a value or tuple of the correct shape. To get a lazy derivative/gradient/Jacobian/Hessian use any one of:
+- `ld = lazyderivative(ab::AbstractBackend, f, xs::Number...)`: returns an operator `ld` for multiplying by the derivative of `f` at `xs`. You can apply the operator by multiplication e.g. `ld * y` where `y` is a number if `f` has a single input, a tuple of the same length as `xs` if `f` has multiple inputs, or an array of numbers/tuples.
```
odow, line 63: Should this be `lazy_derivative` (and below) to match the rest of the `snake_case`?

README.md
mohamed82008: Yes. I am happy to remove this export if folks here agree that it's not useful. Technically, one can use `const` to alias any name in a package but I am not sure if this justifies not exporting anything at all. Exporting makes life easier, that's all. But I don't have strong opinions on this.

README.md
mohamed82008: Valid point. I will open an issue.

Issue #18, opened 2021-10-04
Test suite fails on Julia v1.0


sethaxen:
The fallbacks use a number of idioms (e.g. `eachcol` and `(::UniformScaling)(n)`) that are not present in Julia v1.0.

PR #19, opened 2021-10-04
Add CI workflow

sethaxen:
This PR adds a CI workflow to catch cases like #18


mohamed82008:
It's ok I will do the (little) extra work for JuMP :)

PR #23, opened 2021-10-12
Support complex pullbacks

### LINKED ISSUE #22: Complex pullback support
I realized that if the `_dot` function introduced in #21 is wrapped with `real`, then `pullback_function` works for complex arrays as well, whenever `gradient` supports complex inputs for a given backend. But at least in the scalar case, the real dot product can be computed twice as efficiently as `real(_dot(x, y))`.

sethaxen:
Replaces `_dot` with `_realdot`, which allows for pulling back through functions with complex outputs. Resolves #22. This is currently untested.


mohamed82008:
Could you please add a test here for the intended use case?

sethaxen:
Certainly, though the tests need a few refactors first to make testing these cases more straightforward. Some changes made in #27. More will be necessary.

sethaxen:
Also, this PR should use [RealDot.jl](https://github.com/JuliaMath/RealDot.jl) now that it exists.

PR #24, opened 2021-11-10
Support and test all v1 Julia versions

### LINKED ISSUE #18: Test suite fails on Julia v1.0

mohamed82008:
This PR fixes the support for Julia 1.0, 1.1 and 1.2. The Zygote over ForwardDiff Hessian tests are failing for Zygote mutation-related reasons that seem to be fixed in Julia 1.3 and above. Since this is a Zygote issue for Julia < 1.3, I just avoid testing this case instead of fixing it. This issue fixes #18.


odow:
In general, I think we only need to test the LTS and latest minor release. So just '1.0' and '1'. People on minor releases must update each release. If they don't like that, use the LTS instead.

PR #27, opened 2022-01-09
Make tests more modular

sethaxen:
This PR refactors the tests to be more modular, which will make adding tests for additional backends more straightforward. Some notes:
- each test file is structured so it can be independently `include`d for testing a single component
- the `isa` checks in the test functions that special-cased for certain backends now are changed by keyword arguments
- the test functions and any constants used in them are now declared in `test_utils.jl`.


test/test_utils.jl
```diff
@@ -0,0 +1,394 @@
...
+        # test with respect to analytical solution
+        @test minimum(isapprox.(pf1, (jxvp(xvec,yvec,v[1]), jyvp(xvec,yvec,v[2])), atol=1e-10))
+        @test xvec == xvec2
+        @test yvec == yvec2
+    end
+
+    valvec1, pf1 = AD.value_and_pushforward_function(backend, x -> fjac(x, yvec), xvec)(v[1])
+    valvec2, pf2 = AD.value_and_pushforward_function(backend, y -> fjac(xvec, y), yvec)(v[2])
+
+    # NOTE: added for compatibility with FDMBackend2. Why is this needed?
```
sethaxen, line 185: @mohamed82008 do you know this special-casing is needed?

mohamed82008: it shouldn't be needed

sethaxen: Seems to be though (see latest commit and failing tests).

test/defaults.jl
```diff
@@ -0,0 +1,252 @@
...
+end
+
+struct FDMBackend2{A} <: AD.AbstractFiniteDifference
+    alg::A
+end
+FDMBackend2() = FDMBackend2(central_fdm(5, 1))
+const fdm_backend2 = FDMBackend2()
+AD.@primitive function pushforward_function(ab::FDMBackend2, f, xs...)
+    return function (vs)
+        FDM.jvp(ab.alg, f, tuple.(xs, vs)...)
```
frankschae, line 27: probably, we should add here the conversion to a tuple? (To avoid the conversion in the tests)

sethaxen: Done! Seems to have done the trick.

test/test_utils.jl
```diff
@@ -0,0 +1,390 @@
...
+jyvp(x,y,v) = dfjacdy(x,y)*v
+
+# vJp
+vJxp(x,y,v) = dfjacdx(x,y)'*v
+vJyp(x,y,v) = dfjacdy(x,y)'*v
+
+const xscalar = rand()
+const yscalar = rand()
+
+const xvec = rand(5)
```
frankschae, line 39: I wonder if we should move these `const`s somewhere else (maybe in the runtests file?) to avoid the redefinition warning (since the seed is fixed, they can however probably also stay here).

frankschae (approved):
LGTM!

test/defaults.jl
mohamed82008: agreed

test/test_utils.jl
sethaxen: Moving them to `runtests.jl` would prevent from running each test file independently. What we could do alternatively is make them non-`const`. We could also hard-code them.

sethaxen: @frankschae I opted to move all randomness out of `test_utils.jl`, hardcoding the constants, and allowing the testing functions to accept an RNG. Now each test file is responsible for setting its own seed or providing an RNG.

test/runtests.jl
```diff
@@ -1,665 +1,7 @@
...
-            test_lazy_hessians(backends)
-            if VERSION >= v"1.3"
-                # backends = AD.HigherOrderBackend((zygote_backend1,forwarddiff_backend))
-                # test_lazy_hessians(backends)
-                backends = AD.HigherOrderBackend((zygote_backend1,forwarddiff_backend1))
-                test_lazy_hessians(backends)
-            end
-        end
-    end
+    include("defaults.jl")
```
mohamed82008, line 5: It seems we are including the `test_utils.jl` file twice. Should we perhaps make use of SafeTestsets here or include the file only once?

sethaxen: I realized the workflow I wanted (running just one test file) could be accomplished with
```julia
include("test_utils.jl")
include("forwarddiff.jl")
```
so I now include it only once.

mohamed82008 (approved):
LGTM

mohamed82008:
Thanks @sethaxen !

PR #32, opened 2022-01-19
Support chunks in ForwardDiff backend

devmotion:
I noticed that the ForwardDiff backend doesn't support custom chunk sizes. This PR tries to fix this issue.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AbstractDifferentiation"
 uuid = "c29ec348-61ec-40c8-8164-b8c60e9d9f3d"
 authors = ["Mohamed Tarek <mohamed82008@gmail.com> and contributors"]
-version = "0.2.1"
+version = "0.3.0"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
--- a/src/forwarddiff.jl
+++ b/src/forwarddiff.jl
@@ -1,9 +1,31 @@
 using .ForwardDiff: ForwardDiff, DiffResults, StaticArrays
 
-struct ForwardDiffBackend{CS} <: AbstractBackend
-    ForwardDiffBackend{CS}() where {CS} = new{CS}()
+"""
+    ForwardDiffBackend{CS}
+
+AD backend that uses forward mode with ForwardDiff.jl.
+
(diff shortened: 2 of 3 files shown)


src/forwarddiff.jl
```diff
@@ -1,6 +1,9 @@
 using .ForwardDiff: ForwardDiff, DiffResults, StaticArrays
 
-struct ForwardDiffBackend <: AbstractBackend end
+struct ForwardDiffBackend{CS} <: AbstractBackend
+    ForwardDiffBackend{CS}() where {CS} = new{CS}()
```
mohamed82008, line 4: is this inner constructor necessary?

src/forwarddiff.jl
```diff
@@ -1,6 +1,9 @@
...
+end
+ForwardDiffBackend(; chunksize=nothing) = ForwardDiffBackend{getchunksize(chunksize)}()
```
mohamed82008, line 6: perhaps put a type annotation on `chunksize`?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "AbstractDifferentiation"
 uuid = "c29ec348-61ec-40c8-8164-b8c60e9d9f3d"
 authors = ["Mohamed Tarek <mohamed82008@gmail.com> and contributors"]
-version = "0.2.1"
+version = "0.3.0"
```
sethaxen, line 4: (how) is this a breaking change?

devmotion: The additional type parameter breaks uses of the `ForwardDiffBackend` type, e.g., `SomeOtherType{ForwardDiffBackend}` won't dispatch correctly anymore.

src/forwarddiff.jl
devmotion: Done.

src/forwarddiff.jl
devmotion: No :smile: I removed it.

src/forwarddiff.jl
```diff
@@ -1,6 +1,9 @@
 using .ForwardDiff: ForwardDiff, DiffResults, StaticArrays
 
-struct ForwardDiffBackend <: AbstractBackend end
```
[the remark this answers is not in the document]
devmotion: I added some docstrings.

sethaxen (approved):

Issue #34, opened 2022-01-20
Make value_and_pullback_function a primitive instead of pullback_function


sethaxen:
Currently the internals use closures to avoid computing the primal more than once, but it would be easier to read, simpler, and more consistent with ChainRules, Zygote, Diffractor, etc to instead make `value_and_pullback_function` the primitive. Should come after #4.

Not certain if the same should be done for `pushforward_function`.

Issue #4, opened 2021-08-29
reimplement `value_and_pullback_function` and `value_and_pushforward_function`


mohamed82008:
The current behavior of `value_and_pullback_function`/`value_and_pushforward_function` is that it returns a function which when called with the co-tangents/tangents as input returns both the function value at `x` and the tangent/co-tangent. The delayed return of the value of the function is unnecessary.

oxinabox:
Isn't that the whole point of these functions?

Otherwise would be `value_and_derivative`?
Which is a different function already provided

mohamed82008:
Kind of. Currently the function returned by `value_and_pullback_function` takes a single input (co-tangents) and returns the function value and the pulled back co-tangents. So the value is returned in a lazy fashion after the co-tangents are passed in which is odd.

mohamed82008:
it's not a show-stopper right now as most users won't even use this function

PR #35, opened 2022-01-22
Redefines value_and_pullback_function to return a value and a pullback function

sethaxen:
Fixes part of #4. It's a breaking change, but the version number is given a `-DEV` suffix, since we should also handle #34 before releasing. (EDIT: Ref #36)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AbstractDifferentiation"
 uuid = "c29ec348-61ec-40c8-8164-b8c60e9d9f3d"
 authors = ["Mohamed Tarek <mohamed82008@gmail.com> and contributors"]
-version = "0.3.0"
+version = "0.4.0-DEV"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ This operation goes by a few names. Refer to the [ChainRules documentation](http
 
 The following functions can be used to request the pullback operator/function with or without the function value. In order to request the pullback function `pb_f` of a function `f` at the inputs `xs`, you can use either of:
 - `pb_f = AD.pullback_function(ab::AD.AbstractBackend, f, xs...)`: returns the pullback function `pb_f` of the function `f` at the inputs `xs`. `pb_f` is a function that accepts the co-tangents `vs` as input which is a tuple of length equal to the number of outputs of `f`. If `f` has a single output, `pb_f` can also accept a single input instead of a 1-tuple.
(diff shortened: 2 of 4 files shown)


sethaxen:
@mohamed82008 in https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/36#issuecomment-1021927948
> This PR needs more thinking. I think when using the Jacobian as a primitive, deriving the `value_and_pullback function` won't be possible. Only the `value and pullback_function` is. So it might make sense to define the `value_and_pullback_function/value_and_pushforward_function` to throw when the Jacobian is used as a primitive. That is users shouldn't expect to use these 2 functions for such a backend or at least the methods won't be define automatically.

I'm not sure I understand what you mean. `value_and_pushforward_function` is not touched in this PR. And the tests already include examples where Jacobian is used as a primitive and `value_and_pushforward_function` is tested. e.g. this works with this PR:

```julia
julia> using ForwardDiff, AbstractDifferentiation

julia> struct ForwardDiffBackend1 <: AD.AbstractForwardMode end

julia> AD.primal_value(::ForwardDiffBackend1, ::Any, f, xs) = ForwardDiff.value.(f(xs...))

julia> AD.@primitive function jacobian(ab::ForwardDiffBackend1, f, xs)
           if xs isa Number
               return (ForwardDiff.derivative(f, xs),)
           elseif xs isa AbstractArray
               out = f(xs)
               if out isa Number
                   return (adjoint(ForwardDiff.gradient(f, xs)),)
               else
                   return (ForwardDiff.jacobian(f, xs),)
               end
           elseif xs isa Tuple
               error(typeof(xs))      
           else
               error(typeof(xs)) 
           end
       end

julia> ba = ForwardDiffBackend1();

julia> val, back = AD.value_and_pullback_function(ba, identity, randn(10))
([0.05771077577524417, -0.6046897384683757, -0.43893636554383, 1.008345527744184, -1.3448119234793194, 0.0541982394477022, 0.2593213719269091, 1.6426658474014317, 0.651265342373578, -0.795563384561261], AbstractDifferentiation.var"#21#23"{ForwardDiffBackend1, typeof(identity), Tuple{Vector{Float64}}}(ForwardDiffBackend1(), identity, ([0.05771077577524417, -0.6046897384683757, -0.43893636554383, 1.008345527744184, -1.3448119234793194, 0.0541982394477022, 0.2593213719269091, 1.6426658474014317, 0.651265342373578, -0.795563384561261],)))

julia> back(randn(10))
([-1.2729554614899294, 0.1378713003865544, 0.2690893343964529, 0.9206297829092358, -1.3913076264330144, 0.4862468894828594, -1.1728445800434935, 0.6753224021990761, 0.14583421961218776, 0.732158895744132],)
```

devmotion:
This change would be very helpful for other PRs such as #93: https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/93#discussion_r1279975890

gdalle:
@devmotion should we close this one?

devmotion:
Probably. We can always re-open if it is preferred over #93.

PR #36, opened 2022-01-22
Use value_and_pullback_function as a primitive

### LINKED ISSUE #34: Make value_and_pullback_function a primitive instead of pullback_function

sethaxen:
Fixes #34. Breaking change marked with `-DEV` version number in case it is merged before #35.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AbstractDifferentiation"
 uuid = "c29ec348-61ec-40c8-8164-b8c60e9d9f3d"
 authors = ["Mohamed Tarek <mohamed82008@gmail.com> and contributors"]
-version = "0.3.0"
+version = "0.4.0-DEV"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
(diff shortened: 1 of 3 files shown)


mohamed82008:
This PR needs more thinking. I think when using the Jacobian as a primitive, deriving the `value_and_pullback function` won't be possible. Only the `value and pullback_function` is. So it might make sense to define the `value_and_pullback_function/value_and_pushforward_function` to throw when the Jacobian is used as a primitive. That is users shouldn't expect to use these 2 functions for such a backend or at least the methods won't be define automatically.

mohamed82008:
Ah sorry I meant to comment on the other PR. Anyways, I suggest merging this PR and #34.

sethaxen:
> Ah sorry I meant to comment on the other PR. Anyways, I suggest merging this PR and #34 and updating the README once.

I assume you mean #35? That PR already updates the ReadMe for the redefinition of `value_and_pullback_function`. Currently definition of a primitive is not documented, so we need no updates in this PR.

mohamed82008:
I want to take a stab at this myself if you don't mind. I will open another PR.

gdalle:
@devmotion should we close this one?

devmotion:
Probably. We can always re-open if it is preferred over https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/93.

PR #37, opened 2022-01-22
 Implement backend for FiniteDifferences

sethaxen:
This PR implements a backend for FiniteDifferences.jl. It basically combines the 3 declarations in the tests into 1, with generalization of `pullback_function` to multi-output functions.

Informal benchmarks showed that all three of these methods were necessary for speed. They also showed that with these defined, our default `gradient` and multi-arg `derivative` methods are faster than custom overloads using `jvp` and `grad`. A single-arg `derivative` method was faster than our default, yet for some reason it was less accurate.


mohamed82008:
thanks Seth!

Issue #40, opened 2022-01-26
Planned backends to implement


sethaxen:
We should add backends for the following AD/FD packages:
- [ ] all ChainRules-supporting ADs (see #11, #39)
- [ ] FiniteDiff
- [ ] Tracker
- [ ] Enzyme

AriMKatz:
Can you add Yota also ?

sethaxen:
Yota is ChainRules-compatible, so it should be covered with the others.

wsmoses:
Make sure to add both Enzyme forward and reverse!

sethaxen:
Will do! Should I start with the public API? @frankschae said you had mentioned we might want to use some internal functions (he pointed me to https://github.com/wsmoses/Enzyme.jl/blob/2ce81ffa8f56c5bf44a4d85234c2110fa9d6eb0a/src/compiler.jl#L1745)

wsmoses:
I might not go quite that low level to save yourself some common LLVM setup, but probably using the thunk level (https://github.com/wsmoses/Enzyme.jl/blob/2ce81ffa8f56c5bf44a4d85234c2110fa9d6eb0a/src/compiler.jl#L2700) which has options for "combined" augmented forward pass+gradient, an augmented forward pass (storing values from the original function that need preservation), a standalone gradient (just running the reverse, using the stored values from an augmented forward pass), and forward mode AD.

mohamed82008:
I would like to add a "batch" version of Zygote as a backend which falls back on Zygote except for `jacobian` where the pullback is called with all the bases simultaneously. This can be useful to preserve sparsity of Jacobians if all the rules are written in a way that preserves sparsity.

mohamed82008:
And a `SparseDiffTools` backend to optimise for sparsity structure

sethaxen:
> I would like to add a "batch" version of Zygote as a backend which falls back on Zygote except for `jacobian` where the pullback is called with all the bases simultaneously (i.e. `pb(I)` where `I` is the identity matrix). This can be useful to preserve sparsity of Jacobians if all the rules are written in a way that preserves sparsity.

Is this a feature Zygote actually supports, or just something that sometimes works?

mohamed82008:
>or just something that sometimes works?

Something that sometimes works. The goal is to make it easy to define a sparse Jacobian in a rrule and then get it back when calling `Zygote.jacobian`.

JTaets:
[...]

PR #44, opened 2022-01-26
Implement backend for Tracker

sethaxen:
Relates #40

### DIFF
--- a/src/tracker.jl
+++ b/src/tracker.jl
@@ -1,0 +1,25 @@
+using .Tracker: Tracker
+
+primal_value(x::Tracker.TrackedArray) = Tracker.data(x)
+primal_value(x::Tracker.TrackedReal) = Tracker.data(x)
+primal_value(x::AbstractArray{<:Tracker.TrackedReal}) = Tracker.data.(x)
+
+"""
+    TrackerBackend
+
+AD backend that uses reverse mode with Tracker.jl.
+"""
+struct TrackerBackend <: AbstractReverseMode end
+
+@primitive function pullback_function(ba::TrackerBackend, f, xs...)
+    value, back = Tracker.forward(f, xs...)
+    function pullback(ws)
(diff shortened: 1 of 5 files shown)


sethaxen:
Tracker's an interesting case because I think you want to avoid calling `Tracker.data` on the outputs only if you're nesting the AD (note the Hessian failures). So here we may do something like add a type argument to the backend to indicate whether nesting is happening and then `lowest` would set that to `true`, and based on that we would unwrap the outputs or not.

sethaxen:
Actually, even if we track nesting in the constructor, we still have the problem that the test suite uses `prod` to test Hessians, but Tracker computes the wrong Hessian for `prod`: https://github.com/FluxML/Tracker.jl/issues/120. Perhaps the testing functions should allow the user to provide a function and inputs to be used for testing.

src/tracker.jl
```diff
@@ -0,0 +1,33 @@
...
+
+"""
+    TrackerBackend
+
+AD backend that uses reverse mode with Tracker.jl.
+"""
+struct TrackerBackend <: AbstractReverseMode end
+
+function second_lowest(::TrackerBackend)
+    return throw(ArgumentError("Tracker backend does not support nested differentiation."))
```
sethaxen, line 11: This is the best way I could think of to signal that Tracker should not be used for higher order AD.

sethaxen:
This PR is now ready for review.

Our version of `jacobian` is slower than Tracker's, but it seems to be less brittle. That being said, Tracker as a whole seems pretty brittle (e.g. `Tracker.jacobian(cumsum, x)` fails) and is likely not used much. Still, I think it should be included for now for completeness.

mohamed82008 (approved):
Looks good!

mohamed82008:
Thanks Seth and sorry for the delay.

Issue #39, opened 2022-01-24
Allow arbitrary objects to be backends?


sethaxen:
Currently the fallbacks in the package constrain the backend type to be an `AbstractBackend`. Is this necessary? e.g. as I think @oxinabox proposed on Slack, in relation to #11, it would be nice if a user-created `ChainRules.RuleConfig` could directly be used as a backend. Otherwise, one would probably end up implementing a duplicate object or implementing a loose wrapper around it.

How does this package intend to use these types in a way that can't be satisfied by overloading some interface functions?

mohamed82008:
I thought some more about this and I think a wrapper around a RuleConfig makes more sense than allowing it to be a backend. In particular allowing RuleConfig to be a backend directly will mess with the HigherOrderBackend stuff.

PR #49, opened 2022-02-08
RuleConfig and Zygote support

mohamed82008:
This PR implements RuleConfig support via rrule_via_ad. I tested Zygote and tests pass. I tried Yota but tests didn't pass for some reason so I will open another PR. This is a very minimal PR to get things working but I didn't do the excellent benchmarks @sethaxen does in these kinds of PRs. I think optimisations can be added as a later PR.

### DIFF
--- a/src/ruleconfig.jl
+++ b/src/ruleconfig.jl
@@ -1,5 +1,3 @@
-using ChainRulesCore: RuleConfig, rrule_via_ad
-
 """
     ReverseRuleConfigBackend
 
--- a/src/ruleconfig.jl
+++ b/src/ruleconfig.jl
@@ -1,0 +1,21 @@
+using ChainRulesCore: RuleConfig, rrule_via_ad
+
+"""
+    ReverseRuleConfigBackend
+
+AD backend that uses reverse mode with any ChainRules-compatible reverse-mode AD package.
+"""
+struct ReverseRuleConfigBackend{RC <: RuleConfig} <: AbstractReverseMode
+    ruleconfig::RC
+end
+
+AD.@primitive function pullback_function(ab::ReverseRuleConfigBackend, f, xs...)
+    return (vs) -> begin
+        _, back = rrule_via_ad(ab.ruleconfig, f, xs...)
+        if vs isa Tuple && length(vs) === 1
+            return Base.tail(back(vs[1]))
+        else
+            return Base.tail(back(vs))
+        end
+    end
+end
--- a/Project.toml
+++ b/Project.toml
@@ -17,7 +17,6 @@ ExprTools = "0.1"
 ForwardDiff = "0.10"
 Requires = "0.5, 1"
 ReverseDiff = "1"
-Zygote = "0.6"
 julia = "1"
 
 [extras]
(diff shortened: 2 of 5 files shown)


mohamed82008:
This addresses the motivation behind #39.

mohamed82008:
Seems that the Julia 1.0 tests are failing because ZygoteRuleConfig is only available in a recent version of Zygote which is not compatible with Julia 1.0

mohamed82008:
@odow is JuMP likely to drop 1.0 support any time soon?

src/ruleconfig.jl
```diff
@@ -0,0 +1,21 @@
+using ChainRulesCore: RuleConfig, rrule_via_ad
```
sethaxen, line 1: It's not a big deal, but I prefer imports to be in the main module file.

sethaxen (approved):
Perhaps add a patch version bump so we can release after this?

src/ruleconfig.jl
mohamed82008: done

mohamed82008:
RuleConfig support is now available only in Julia 1.6+. I will go ahead and merge this then if 1.0 support is still required for RuleConfig, we can open another issue/PR discussing/implementing that.

src/ruleconfig.jl
```diff
@@ -0,0 +1,19 @@
...
+
+AD backend that uses reverse mode with any ChainRules-compatible reverse-mode AD package.
+"""
+struct ReverseRuleConfigBackend{RC <: RuleConfig} <: AbstractReverseMode
+    ruleconfig::RC
+end
+
+AD.@primitive function pullback_function(ab::ReverseRuleConfigBackend, f, xs...)
+    return (vs) -> begin
+        _, back = rrule_via_ad(ab.ruleconfig, f, xs...)
```
devmotion, line 12: What's reason for moving this in the function body? Isn't it better to obtain `back` only once outside of the function instead of doing it at every invocation of the pullback function?

src/ruleconfig.jl
```diff
@@ -0,0 +1,19 @@
...
+        if vs isa Tuple && length(vs) === 1
```
devmotion, line 13: Does this compile away in the same way as, I would assume, `if vs isa Tuple{Any}`?

src/ruleconfig.jl
```diff
@@ -0,0 +1,19 @@
...
+            return Base.tail(back(vs[1]))
```
devmotion, line 14: It seems this rules out support for functors `f` as it removes the derivatives wrt to `f`? But maybe generally the design of AbstractDifferentiation does not support them?

sethaxen: Yes, this is so. AbstractDifferentiation supports a strictly smaller set of applications than, say, Zygote. It's all about scalars and arrays. And even there, I'm pretty sure results will be inconsistent if one tries to, say, take the gradient of a structured array. And functors are not supported.

src/ruleconfig.jl
sethaxen: This is a good point.

PR #50, opened 2022-02-08
Yota support

mohamed82008:
This PR is based on #49 but tests here fail. I am opening it in case @dfdx or someone else can figure out why they fail when Zygote passes.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,6 +9,7 @@ Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 ExprTools = "e2ba6199-217a-4e67-a87a-7c52f15ade04"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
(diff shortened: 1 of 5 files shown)


dfdx:
I see one error coming from a Yota bug that is already fixed on `main`, but I can't immediately understand the others, especially related to the arity of functions, e.g.:

>   MethodError: no method matching fder(::Float64)
>  Closest candidates are:
>    fder(::Any, ::Any) at /home/runner/work/AbstractDifferentiation.jl/AbstractDifferentiation.jl/test/test_utils.jl:6

Do you have a minimal code example that reproduces them? If not, give me a couple of days to learn internals of AbstractDifferentiation and find where the problem comes from.

mohamed82008:
>Do you have a minimal code example that reproduces them?

Unfortunately no. I can dig into it next week if you don't beat me to it.

dfdx:
I managed to fix a few cases, but the amount of errors only grows, and I don't understand the code base here well enough to understand what exactly is expected. Is there a minimal set of interfaces that need to be implemented? Perhaps, `value_and_xxx` set of methods?

mohamed82008:
Some of the internals here will be redesign soon so I would wait for that to happen before trying to understand them :)

PR #51, opened 2022-02-09
Retrieve pullback only once for ReverseRuleConfigBackend

devmotion:
This PR contains the suggestions in https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/49#pullrequestreview-877962490.

Most importantly, the CR pullback function is only computed once and reused. Moreover, the type parameter is restricted to `RuleConfig`s with reverse mode capability, in line with the intention in the docstring.


mohamed82008 (approved):
Looks good.

mohamed82008:
Thanks David!

PR #52, opened 2022-02-09
Improve CI

devmotion:
This PR
- makes sure that CI runs only once in a PR and is not triggered both by the `push` and `pull_request` event (as e.g. in https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/51)
- cancels CI runs of outdated commits in PRs
- switches to https://github.com/julia-actions/cache which is a) simpler and b) should be faster since packages are cached by default

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,7 +1,16 @@
 name: CI
+
 on:
-  - push
-  - pull_request
+  push:
+    branches: [master]
+  pull_request:
+
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: only if it is a pull request build.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
 jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -1,10 +1,19 @@
 name: CI
+
 on:
-  - push
-  - pull_request
+  push:
+    branches: [master]
```
sethaxen, line 5: Perhaps we should also run on tags?
```suggestion
    branches: [master]
    tags: [v*]
```

.github/workflows/CI.yml
```diff
@@ -1,10 +1,19 @@
...
+  pull_request:
+
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: only if it is a pull request build.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
```
sethaxen, line 12: Ah, I hadn't seen this done before. I assume you've seen it in action and it works as documented.

devmotion: Yeah, it's used all over the place (e.g. ChainRulesCore: https://github.com/JuliaDiff/ChainRulesCore.jl/blob/3e11e627b5a10a3149902b2c538be63c61ecafbc/.github/workflows/CI.yml#L10-L14) and part of PkgTemplates (https://github.com/invenia/PkgTemplates.jl/blob/1cb56bf90326d47c402af51537c9c669e9b08a24/templates/github/workflows/CI.yml#L10-L14).

.github/workflows/CI.yml
devmotion: I don't mind. But I don't think it is necessary - it is run for every commit pushed to the master branch anyway. I think the `tags` trigger is mainly useful for building release-specific documentation.

sethaxen: We'll want that eventually anyways, once we have documentation. I'll leave it up to you whether to add it.

sethaxen (approved):

.github/workflows/CI.yml
devmotion: For documentation we might want to add a separate action anyway, IMO that's cleaner (and e.g. allows more fine-tuned triggers) :shrug:

devmotion: I don't add the trigger in this PR, we can add it later if it seems useful.

Issue #31, opened 2022-01-11
Better description of how to use and/or simple examples needed


alecloudenback:
As a unfamiliar user, I am left confused on how to use this package without examples or more documentation.

> To use AbstractDifferentiation, first construct a backend instance `ab::AD.AbstractBackend` using your favorite differentiation package in Julia that supports AbstractDifferentiation.

How does one do this? Is my favorite differentiation package supposed to document this?

Also, some simple examples (ie basic polynomial functions, not complex ODEs) would help new users by to showing how to use this package.

gdalle:
I'm not an AbstractDifferentiation.jl contributor, but from what I understand, the library defines several backend objects that are made available if the corresponding package is loaded (thanks to the magic of [Requires.jl](https://github.com/JuliaPackaging/Requires.jl)).
Here's an example:

```julia
julia> using AbstractDifferentiation

julia> ab = AD.ZygoteBackend()
ERROR: UndefVarError: ZygoteBackend not defined
Stacktrace:
 [1] getproperty(x::Module, f::Symbol)
   @ Base ./Base.jl:35
 [2] top-level scope
   @ REPL[3]:1

julia> using Zygote

julia> ab = AD.ZygoteBackend()
AbstractDifferentiation.ReverseRuleConfigBackend{Zygote.ZygoteRuleConfig{Zygote.Context}}(Zygote.ZygoteRuleConfig{Zygote.Context}(Zygote.Context(nothing)))

julia> f(x) = log(sum(exp, x))
f (generic function with 1 method)

julia> AD.gradient(ab, f, rand(10))
([0.07163448353282538, 0.08520350535348796, 0.09675622487503996, 0.1522744408520505, 0.12174662595572318, 0.07996969757526722, 0.07832665607158593, 0.11001685581681672, 0.06691909637037166, 0.1371524135968315],)
```

From what I can tell, the following backends are defined by AbstractDifferentiation.jl:

- `AD.TrackerBackend()` for [Tracker.jl](https://github.com/FluxML/Tracker.jl)
- `AD.ReverseDiffBackend()` for [ReverseDiff.jl](https://github.com/JuliaDiff/ReverseDiff.jl)
- `AD.ForwardDiffBackend()` for [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl)
[...]

PR #56, opened 2022-03-27
More details on backends in README

gdalle:
Add explanation about backends and small example (see #31)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,42 @@ using AbstractDifferentiation
 
 ## `AbstractDifferentiation` backends
 
-To use `AbstractDifferentiation`, first construct a backend instance `ab::AD.AbstractBackend` using your favorite differentiation package in Julia that supports `AbstractDifferentiation`. For higher order derivatives, you can build higher order backends using `AD.HigherOrderBackend`. For instance, let `ab_f` be a forward-mode automatic differentiation backend and let `ab_r` be a reverse-mode automatic differentiation backend. To construct a higher order backend for doing forward-over-reverse-mode automatic differentiation, use `AD.HigherOrderBackend((ab_f, ab_r))`. To construct a higher order backend for doing reverse-over-forward-mode automatic differentiation, use `AD.HigherOrderBackend((ab_r, ab_f))`.
+To use `AbstractDifferentiation`, first construct a backend instance `ab::AD.AbstractBackend` using your favorite differentiation package in Julia that supports `AbstractDifferentiation`.
+The following backends are made available by `AbstractDifferentiation` if their corresponding package is loaded (thanks to [Requires.jl](https://github.com/JuliaPackaging/Requires.jl)):
+
+- `AD.ForwardDiffBackend()` for [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl)
+- `AD.FiniteDifferencesBackend()` for [FiniteDifferences.jl](https://github.com/JuliaDiff/FiniteDifferences.jl)
+- `AD.ReverseDiffBackend()` for [ReverseDiff.jl](https://github.com/JuliaDiff/ReverseDiff.jl)
+- `AD.ReverseRuleConfigBackend(ruleconfig)` for any [ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl)-compatible reverse mode backend
+- `AD.ZygoteBackend()` for [Zygote.jl](https://github.com/FluxML/Zygote.jl)
+- `AD.TrackerBackend()` for [Tracker.jl](https://github.com/FluxML/Tracker.jl)
+
+Here's an example:
+
+```julia
+julia> using AbstractDifferentiation
+
+julia> ab = AD.ZygoteBackend()
+ERROR: UndefVarError: ZygoteBackend not defined
+Stacktrace:
+ [1] getproperty(x::Module, f::Symbol)
+   @ Base ./Base.jl:35
+ [2] top-level scope
+   @ REPL[3]:1
+
+julia> using Zygote
+
+julia> ab = AD.ZygoteBackend()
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,7 +19,42 @@ using AbstractDifferentiation
 
 ## `AbstractDifferentiation` backends
 
-To use `AbstractDifferentiation`, first construct a backend instance `ab::AD.AbstractBackend` using your favorite differentiation package in Julia that supports `AbstractDifferentiation`. For higher order derivatives, you can build higher order backends using `AD.HigherOrderBackend`. For instance, let `ab_f` be a forward-mode automatic differentiation backend and let `ab_r` be a reverse-mode automatic differentiation backend. To construct a higher order backend for doing forward-over-reverse-mode automatic differentiation, use `AD.HigherOrderBackend((ab_f, ab_r))`. To construct a higher order backend for doing reverse-over-forward-mode automatic differentiation, use `AD.HigherOrderBackend((ab_r, ab_f))`.
+To use `AbstractDifferentiation`, first construct a backend instance `ab::AD.AbstractBackend` using your favorite differentiation package in Julia that supports `AbstractDifferentiation`.
+The following backends are made available by `AbstractDifferentiation` if their corresponding package is loaded (thanks to [Requires.jl](https://github.com/JuliaPackaging/Requires.jl)):
```
devmotion, line 23: FYI this is only a workaround until the API is stabilized. Ideally these backends will be defined in the respective AD packages.

README.md
```diff
@@ -19,7 +19,42 @@ using AbstractDifferentiation
...
+
+- `AD.ForwardDiffBackend()` for [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl)
+- `AD.FiniteDifferencesBackend()` for [FiniteDifferences.jl](https://github.com/JuliaDiff/FiniteDifferences.jl)
+- `AD.ReverseDiffBackend()` for [ReverseDiff.jl](https://github.com/JuliaDiff/ReverseDiff.jl)
+- `AD.ReverseRuleConfigBackend(ruleconfig)` for any [ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl)-compatible reverse mode backend
```
devmotion, line 28: This is always available and not dependent on Requires. `ZygoteBackend` is just a special variant of it, with `ZygoteRuleConfig` as `ruleconfig` (as you can see below). For other CR-compatible AD backends one would use their respective rule configs.

README.md
```diff
@@ -19,7 +19,42 @@ using AbstractDifferentiation
...
+julia> using AbstractDifferentiation
+
+julia> ab = AD.ZygoteBackend()
+ERROR: UndefVarError: ZygoteBackend not defined
+Stacktrace:
+ [1] getproperty(x::Module, f::Symbol)
+   @ Base ./Base.jl:35
+ [2] top-level scope
+   @ REPL[3]:1
+
```
devmotion, line 44: I'm not sure if this is helpful. Maybe sufficient to just show how it works?
```suggestion
```

README.md
gdalle: Good to know! In the meantime I still think it can help users to know what's available without digging into the source code of AbstractDifferentiation.jl as I did.

README.md
gdalle: Of course it is not dependent on Requires being loaded. What I meant is that AbstractDifferentation uses Requires to provide this in a transparent manner (see [here](https://github.com/JuliaDiff/AbstractDifferentiation.jl/blob/12a0e27d0f54e6300b725dfb9dafc21ef507b724/src/AbstractDifferentiation.jl#L648)).

As an AD novice, I'm not yet sure what a `RuleConfig` is, so I guess it wouldn't hurt to add an example of that too?

README.md
gdalle: I think it helps to underline the fact that AbstractDifferentiation won't load your package for you. Again, speaking from a beginner's perspective, this is something that I would love to know right away, and this example shows it clearly.

README.md
devmotion: Sure, I just wanted to point out that this will change at some point and Requires will be dropped then.

README.md
devmotion: No, my point is that this backend does not use Requires. It is always available, even if you don't load any AD package.

`RuleConfig` is a CR thing: https://juliadiff.org/ChainRulesCore.jl/stable/rule_author/superpowers/ruleconfig.html
For AD packages that are compatible with CR an AD-specific `RuleConfig` should be defined: https://juliadiff.org/ChainRulesCore.jl/stable/rule_author/superpowers/ruleconfig.html#Writing-rules-that-are-only-for-your-own-AD E.g., Zygote defines a `ZygoteRuleConfig` `RuleConfig` that allows to call (back) into the Zygote AD pipeline.

You already have an example here, since `ZygoteBackend() == ReverseRuleConfigBackend(Zygote.ZygoteRuleConfig())` :smile:

README.md
devmotion: Sure, but isn't documenting this and showing the correct code sufficient? It still seems strange to me to include a failing example here. I think I don't know any package with such a Requires-setup that does show how to not do it.

README.md
gdalle:
```suggestion
The following backends are temporarily made available by `AbstractDifferentiation` as soon as their corresponding package is loaded (thanks to [Requires.jl](https://github.com/JuliaPackaging/Requires.jl)).
In the long term, these backend objects are meant be defined within their respective packages to enforce the `AbstractDifferentiation` interface.
```

README.md
gdalle: Ah, right, I hadn't understood that the first part of the sentence referred to `ReverseRuleConfigBackend` since the whole list was highlighted, my bad :blush:

README.md
gdalle: Your call. I just thought that it may help users unfamiliar with Requires, but maybe it does more harm than good indeed

devmotion (approved):
LGTM, thanks!

mohamed82008:
Thanks @gdalle and sorry for the super late merge!

mohamed82008:
Thanks @devmotion!

Issue #33, opened 2022-01-20
Function as first argument?


devmotion:
Is there a specific reason for making the function `f` the second argument? It seems a bit unfortunate that this makes it impossible to use the `do` syntax which one could use if `f` would be the first argument such as in ForwardDiff or Zygote (as suggested also by the [Julia style guide](https://docs.julialang.org/en/v1/manual/style-guide/index.html#Write-functions-with-argument-ordering-similar-to-Julia-Base)).

sethaxen:
`gradient(ba, f, args...)` just looks cleaner than `gradient(f, ba, args...)`, since it puts the functions with its args. Also, in general these functions should dispatch on the backend, not `f`, so it's cleaner it be the first argument. There's other precedence for this as well in AD packages (see e.g. [`ChainRulesCore.rrule_via_ad`](https://juliadiff.org/ChainRulesCore.jl/stable/api.html#ChainRulesCore.rrule_via_ad)).

Zygote is able to get away with this since the user never needs to provide a config object. Same with ForwardDiff.

But it would be better to support `do` syntax. To that point, I wonder if it would make sense to support alternate syntax for the API functions, where the backend is a keyword argument. e.g.
```julia
gradient(f, xs...; backend) = gradient(backend, f, xs...)
```

marius311:
Implemented @sethaxen's suggestion in https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/62

phipsgabler:
What about a curried form?

```julia
gradient(ab::AbstractBackend) = FixBackend(ab, gradient)
(op::FixBackend)(f, xs...) = op.∂(op.backend, f, xs...)
```

Then one can write

```julia
gradient(SomeBackend())(a, b) do x, y
    ...
end

sethaxen:
@phipsgabler what do you see as the pros and cons of the curried form vs the one in https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/33#issuecomment-1018923309?

phipsgabler:
Purely subjectively, I like the looks of it better. The backend argument looks like something that should be more privileged than a mere "option hidden in a kwarg", IMHO.

Also (and this would be a question practitioners): isn't fixing the backend and then using the same differential operator repeatedly within a package a common use case?

devmotion:
[...]

PR #62, opened 2022-08-18
add do-block-friendly forms

marius311:
I also really want do-block with this package and the suggestion here https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/33#issuecomment-1018923309 seems like the best one to me, so I implemented it in this PR should the devs like it. 

For reference, I think the other options are 

1. switch the order and break everything (:vomiting_face:) 
2. allow both orders with dispatch (breaks the only-one-way-to-do-something rule, which I think is worse than this PR but honestly only very marginally worse, so maybe worth considering, I'm not married to this PR, I just want do-block :grin: )

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,14 @@ julia> AD.gradient(ab, f, rand(10))
 ([0.07163448353282538, 0.08520350535348796, 0.09675622487503996, 0.1522744408520505, 0.12174662595572318, 0.07996969757526722, 0.07832665607158593, 0.11001685581681672, 0.06691909637037166, 0.1371524135968315],)
 ```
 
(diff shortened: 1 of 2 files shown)


marius311:
Fun fact, Github Copilot did most of the work :sweat_smile: 

image

marius311:
Btw I think the test error is unrelated. Is there anything else I could do to help get this merged or decide on a better option?

marius311:
Any interest in this PR? Just reran CI, looks good. 

To address the idea from https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/33#issuecomment-1229150339, IMO a curried do-block is less standard (I don't think I've seen it anywhere in Julia yet). As for fixing the backend, I think this style also gets you that, with a succinct, standard, and readable separation between arguments and backend:

```julia
# somewhere high up
backend = ...

# throughout:
gradient(x, y, z; backend) do x, y, z
    ...
end
```

devmotion:
As I mentioned in https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/33#issuecomment-1361169696, I am leaning towards the curried form.

marius311:
Closing as seems not the preferred solution.

PR #68, opened 2022-12-27
Use weak dependencies if supported

devmotion:
This PR replaces the use of Requires with weak dependencies on Julia >= 1.9 (https://pkgdocs.julialang.org/dev/creating-packages/#Weak-dependencies).

Currently the PR is breaking since it removes the `AD` alias and export. I couldn't get the Requires/weak dependencies submodules working with the `AD` alias on a first try (I tried different combinations of `using AbstractDifferentiation: AD`, `using AbstractDifferentiation: AbstractDifferentiation`, `using AD: AD` etc.). I'm sure there's some way to fix it but I did not spend more time on it since
- my impression was it is not really needed anyways due to the ["renaming with `as`"](https://docs.julialang.org/en/v1/manual/modules/#Renaming-with-as) feature in Julia >= 1.6,
- the problems seemed to indicate its not an intended/common use case (actually, the only where I've seen this pattern is AbstractDifferentiation - e.g., FiniteDifferences only mentions and uses the alias `FDM` in the docs but does not define it in the package itself)

~~Edit: Shouldn't be merged until the Registrator deployment of https://github.com/JuliaRegistries/Registrator.jl/pull/397 is done.~~

Edit: A few additional changes (see below):
 - As suggested by @ChrisRackauckas in https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/68#discussion_r1108508383, I also used the opportunity to simplify how/when Zygote is loaded and bumped the Julia compat to 1.6.
- As mentioned in https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/68#issuecomment-1448234084, the macros made too strong assumptions about what is available in the current scope but are fixed now
- When updating the julia compat to 1.6, I also replaced a few `@assert length(x) == 1; x[1]` with `only(x)`


mohamed82008:
This is a great PR, thanks! I need to study this feature before I can review the PR. But if someone else approves the PR, I am happy to merge.

devmotion:
Maybe in addition to the Pkg docs linked above, it could be interesting to check Kristoffer's PR to PGFPlotsX: https://github.com/KristofferC/PGFPlotsX.jl/pull/306

ChrisRackauckas:
Can you quasi-namespace the extensions? I.e. AbstractDifferentiationForwardDiffExt? Otherwise you'll get collisions.

src/AbstractDifferentiation.jl
```diff
@@ -645,14 +644,21 @@ end
...
+if !EXTENSIONS_SUPPORTED
+    using Requires: @require
+end
 function __init__()
-    @require ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210" include("forwarddiff.jl")
-    @require ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267" include("reversediff.jl")
-    @require FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000" include("finitedifferences.jl")
-    @require Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c" include("tracker.jl")
-    @require Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f" begin
+    @static if !EXTENSIONS_SUPPORTED
```
ChrisRackauckas, line 655: can move this out of the init definition

src/AbstractDifferentiation.jl
```diff
@@ -645,14 +644,21 @@ end
...
+        @require ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210" include("../ext/ForwardDiffExt.jl")
+        @require ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267" include("../ext/ReverseDiffExt.jl")
+        @require FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000" include("../ext/FiniteDifferencesExt.jl")
+        @require Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c" include("../ext/TrackerExt.jl")
         @static if VERSION >= v"1.6"
-            ZygoteBackend() = ReverseRuleConfigBackend(Zygote.ZygoteRuleConfig())
+            @require Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f" include("../ext/ZygoteExt.jl")
```
ChrisRackauckas, line 661: why is this Zygote one versioned off?

ChrisRackauckas: For simplicity, just bump to LTS?

devmotion:
> Can you quasi-namespace the extensions? I.e. AbstractDifferentiationForwardDiffExt? Otherwise you'll get collisions.

Sure, I've seen this discussion in other places and Kristoffer said that probably it would be reasonable even if the collision issues are fixed upstream. It's a bit hacky though IMO since just adding a prefix does not guarantee that there won't be any collisions - e.g., an extension such as `ForwardDiffChainRulesStaticArraysExt` could be both an extension of ForwardDiff with weak dependencies ChainRules and StaticArrays or an extension of ForwardDiffChainRules with weak dependency StaticArrays. In any case, I'll update the PR :slightly_smiling_face:

src/AbstractDifferentiation.jl
devmotion: I tried many different versions and this was the (only?) one that seemed to work. But I'll give it another shot.

ChrisRackauckas: https://github.com/SciML/RecursiveArrayTools.jl/blob/master/src/RecursiveArrayTools.jl#L45-L50 it works here?

devmotion: Ah, I think the problem was that I had put `using Requires` in the same check (outside of ìnit) which does not work. Having to `if` statements seems to work though.

devmotion: Actually, I just managed to reproduce the problem: https://github.com/tpapp/LogDensityProblemsAD.jl/actions/runs/4226767173/jobs/7340561574 failed with https://github.com/tpapp/LogDensityProblemsAD.jl/actions/runs/4226767173/jobs/7340561574

devmotion:
Well, well. Updating to `import AbstractDifferentiation as AD` in the extensions revealed that the macros for defining pullbacks etc. have to be fixed: They implicitly assume that `AbstractDifferentiation` is available in the current scope (under this name) and they can't deal with arguments without names etc. So it seems this PR will get quite big, and possibly one even wants to fix a few things on the master branch.

devmotion:
To summarize the last commits:
- As suggested by @ChrisRackauckas in https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/68#discussion_r1108508383, I also used the opportunity to simplify how/when Zygote is loaded and bumped the Julia compat to 1.6.
- As mentioned in https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/68#issuecomment-1448234084, the macros made too strong assumptions about what is available in the current scope but are fixed now
- When updating the julia compat to 1.6, I also replaced a few `@assert length(x) == 1; x[1]` with `only(x)`

The PR is ready for a thorough review and updated with the current practices for defining extensions.

ext/AbstractDifferentiationChainRulesCoreExt.jl
```diff
@@ -0,0 +1,13 @@
+module AbstractDifferentiationChainRulesCoreExt
+
+import AbstractDifferentiation as AD
+using ChainRulesCore: ChainRulesCore
```
ChrisRackauckas, line 4: This is not being imported using the `..ChainRulesCore` that's needed for Requires

src/AbstractDifferentiation.jl
```diff
@@ -644,16 +620,23 @@ end
...
-    @require Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f" begin
-        @static if VERSION >= v"1.6"
-            ZygoteBackend() = ReverseRuleConfigBackend(Zygote.ZygoteRuleConfig())
+include("backends.jl")
+
+# TODO: Replace with proper version
+const EXTENSIONS_SUPPORTED = isdefined(Base, :get_extension)
+if !EXTENSIONS_SUPPORTED
+   using Requires: @require
+   include("../ext/AbstractDifferentiationChainRulesCoreExt.jl")
```
ChrisRackauckas, line 629: I see, because there's no backwards compat to do? It's a bit odd but fine I guess.

ChrisRackauckas (approved):
This looks correct.

ext/AbstractDifferentiationChainRulesCoreExt.jl
devmotion: It's not loaded with Requires.

src/AbstractDifferentiation.jl
devmotion: I followed the same pattern as in other packages: pre-Julia 1.9 ChainRulesCore is a proper dependency (no Requires involved) whereas in >= 1.9 it's an extension.

PR #70, opened 2023-01-05
reset Zygote context cache after each call

### LINKED ISSUE #69: Zygote context cache incorrectly(?) persists between AD calls

```julia
julia> let
           ad = AD.ZygoteBackend()
           AD.derivative(ad, exp, 1.)
           AD.derivative(ad, exp, 1.)
       end
```

marius311:
Fixes #69

I did the reset after to allow GC immediately. Perhaps there's better ways, happy to take comments.

### DIFF
--- a/src/AbstractDifferentiation.jl
+++ b/src/AbstractDifferentiation.jl
@@ -653,6 +653,10 @@ function __init__()
     @require Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f" begin
         @static if VERSION >= v"1.6"
             ZygoteBackend() = ReverseRuleConfigBackend(Zygote.ZygoteRuleConfig())
+            function empty_cache!(ruleconfig::Zygote.ZygoteRuleConfig)
+                ruleconfig.context.cache = nothing
+                ruleconfig
+            end
         end
     end
 end
--- a/src/ruleconfig.jl
+++ b/src/ruleconfig.jl
@@ -9,7 +9,13 @@ end
 
 AD.@primitive function pullback_function(ab::ReverseRuleConfigBackend, f, xs...)
     _, back = ChainRulesCore.rrule_via_ad(ab.ruleconfig, f, xs...)
-    pullback(vs) = Base.tail(back(vs))
-    pullback(vs::Tuple{Any}) = Base.tail(back(first(vs)))
+    function pullback(vs)
+        grad = Base.tail(back(vs))
+        empty_cache!(ab.ruleconfig)
+        grad
+    end
+    pullback(vs::Tuple{Any}) = pullback(first(vs))
     return pullback
 end
+
+empty_cache!(ruleconfig::ChainRulesCore.RuleConfig) = ruleconfig


ChrisRackauckas (approved):
This definitely seems like the correct thing to do.

devmotion:
I wonder if this is a Zygote issue that should be fixed in Zygote. And more generally, if it is OK at all to mutate RuleConfigs in rrule/frule calls.

mohamed82008:
I agree with David. This doesn't seem AbstractDifferentiation specific at all.

marius311:
I think this has to be here since we are calling into `ChainRulesCore.rrule_via_ad`, but you can't clear the cache anywhere inside that function or what it calls since that has no guarantees not to be nested in fact often is; `AD.pullback_function` is the best indication this is a top-level call, ie, its the place to clear the cache.

marius311:
Independent of whether the fix belongs here or in Zygote, marking as draft for now while I think through some consequences of clearing the cache after each jacobian column (which is what this PR currently does) brought up by @touchesir on Slack.

ToucheSir:
> And more generally, if it is OK at all to mutate RuleConfigs in rrule/frule calls.

Short of advanced compiler magic which doesn't exist for pure Julia ADs yet, not sure there's a way around storing mutable state in RuleConfigs if one wants to support `setfield!`.

For the issue at hand, one solution could be to give `ZygoteBackend` its own `@primitive` definitions which create a fresh RuleConfig on each call. That would avoid any jacobian pullback reuse issues and keep any Zygote-specific modifications contained within the `ZygoteBackend` source.

marius311:
Thanks for thinking about this again @ToucheSir. Would be great to get #69 fixed in some way. I'm pretty sure this is above my paygrade though unfortunately, so can't really say I expect to be able to contribute here further. 

For posterity at least here's the example you gave on Slack that convinced me this PR is wrong:

```julia
julia> using Zygote

julia> function f(x, a)
           r = Ref(x)
           r[] = r[] + r[]
           r[] = r[] * a
           r[]
       end
f (generic function with 1 method)

julia> Zygote.gradient(f, 1, 3)
(6.0, 2.0)

julia> Zygote.withjacobian(f, [1 2 3], 3)
(val = [6, 12, 18], grad = ([6 0 0; 0 6 0; 0 0 6], [2, 4, 6]))

julia> function withjacobian(f, args...) # same behaviour as Zygote.withjacobian, just with cache clearing after each pullback call
         cx = Zygote.Context()
         y, back = Zygote.pullback(Zygote._jvec∘f, cx, args...)
         out = map(args) do x
           T = promote_type(eltype(x), eltype(y))
           dx = x isa AbstractArray ? similar(x, T, length(y), length(x)) :
             x isa Number ? similar(y, T, length(y)) :
             nothing
         end
         delta = Zygote._eyelike(y)
         for k in LinearIndices(y)
           grads = back(delta[:,k])
           for (dx, grad) in zip(out, grads)
             dx isa AbstractArray || continue
             Zygote._gradcopy!(view(dx,k,:), grad)
           end
           @show cx
           empty!(cx.cache)
         end
         (val=y, grad=out)
       end
withjacobian (generic function with 1 method)

julia> withjacobian(f, [1 2 3], 3)
cx = Zygote.Context{false}(IdDict{Any, Any}(Base.RefValue{Matrix{Int64}}([6 12 18]) => Base.RefValue{Any}((x = nothing,))))
cx = Zygote.Context{false}(IdDict{Any, Any}(Base.RefValue{Matrix{Int64}}([6 12 18]) => Base.RefValue{Any}((x = [0 1 0],))))
cx = Zygote.Context{false}(IdDict{Any, Any}(Base.RefValue{Matrix{Int64}}([6 12 18]) => Base.RefValue{Any}((x = [0 0 1],))))
(val = [6, 12, 18], grad = ([6 0 0; 0 0 0; 0 0 0], [2, 0, 0]))
```

The `function withjacobian` definition basically emulates this PR, but gives an incorrect grad at the end.

ToucheSir:
It should be a relatively small change. I'd be happy to file a PR once #68 lands.

ChrisRackauckas:
the plane landed.

PR #76, opened 2023-03-03
Don't persist Zygote backend cache

### LINKED ISSUE #69: Zygote context cache incorrectly(?) persists between AD calls

ToucheSir:
Fixes #69 and closes #70.

### DIFF
--- a/ext/AbstractDifferentiationZygoteExt.jl
+++ b/ext/AbstractDifferentiationZygoteExt.jl
@@ -7,6 +7,10 @@ else
     using ..Zygote: Zygote
 end
 
-AD.ZygoteBackend() = AD.ReverseRuleConfigBackend(Zygote.ZygoteRuleConfig())
+AD.@primitive function pullback_function(ba::AD.ZygoteBackend, f, xs...)
+    @inline # @primitive doesn't support this before function keyword
+    rc_backend = AD.ReverseRuleConfigBackend(Zygote.ZygoteRuleConfig())
+    return AD.pullback_function(rc_backend, f, xs...)
+end
 
 end # module
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -71,4 +71,4 @@ It is a special case of [`ReverseRuleConfigBackend`](@ref).
 !!! note
     To be able to use this backend, you have to load Zygote.
 """
-function ZygoteBackend end
+struct ZygoteBackend <: AbstractReverseMode end


ext/AbstractDifferentiationZygoteExt.jl
```diff
@@ -7,6 +7,10 @@ else
     using ..Zygote: Zygote
 end
 
-AD.ZygoteBackend() = AD.ReverseRuleConfigBackend(Zygote.ZygoteRuleConfig())
+AD.@primitive function pullback_function(ba::AD.ZygoteBackend, f, xs...)
+    @inline # @primitive doesn't support this before function keyword
```
devmotion, line 11: This syntax is not supported on Julia 1.6.

devmotion:
```suggestion
```

src/backends.jl
```diff
@@ -71,4 +71,4 @@ It is a special case of [`ReverseRuleConfigBackend`](@ref).
 !!! note
     To be able to use this backend, you have to load Zygote.
 """
-function ZygoteBackend end
+struct ZygoteBackend <: AbstractReverseMode end
```
devmotion, line 74: This also requires changing the docstring and the README which both state that it is just a special `ReverseRuleConfigBackend`.

devmotion (commented):
Could we overload `AD.pullback_function(::ReverseRuleConfig{<:Zygote.ZygoteRuleConfig}, f, xs...)` instead of making `ZygoteRuleConfig` a separate backend? I think it is very unfortunatem and unintuitive if `ReverseRuleBackend(ZygoteRuleConfig())` is broken and behaves differently from `ZygoteBackend`.

ext/AbstractDifferentiationZygoteExt.jl
ToucheSir: That's odd, I thought I force-pushed a new commit which removed it. Either way, #77 is the way to go.

ToucheSir:
I thought about that, but somehow got it in my mind that it would require an ugly `@invoke` to the more general `ReverseRuleConfigBackend` method. The code duplication in #77 is unfortunate but pretty harmless, so closing this in favour of that.

PR #77, opened 2023-03-03
New Zygote context in every call to `AD.pullback_function`

### LINKED ISSUE #69: Zygote context cache incorrectly(?) persists between AD calls

devmotion:
Alternative to #76 that I had in mind when writing https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/76#pullrequestreview-1323649955.

In contrast to #76, with this PR `ZygoteBackend` would still be a `ReverseRuleConfigBackend`, and hence the behaviour of `ZygoteBackend` and `ReverseRuleConfigBackend(Zygote.ZygoteRuleConfig())` would not differ.

Fixes #69. Closes #70. Closes #76.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AbstractDifferentiation"
 uuid = "c29ec348-61ec-40c8-8164-b8c60e9d9f3d"
 authors = ["Mohamed Tarek <mohamed82008@gmail.com> and contributors"]
-version = "0.5.0"
+version = "0.5.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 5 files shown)


devmotion:
I made the fix more modular in https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/77/commits/848ee3847c364bd95f4a0d6a57c7f6787cb3d43f, to avoid code duplication and make it easy to apply similar fixes for other AD backends if needed.

ToucheSir:
Well, that eliminates my only substantive point in https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/76#issuecomment-1453695860. As the adage goes, the best way to get something done is to do it wrong and get someone to correct it for you :laughing:

devmotion:
That is, you approve this PR? 🙂

ToucheSir (approved):
Yes, though I'm not sure how much weight it carries here :)

devmotion:
I trust your Zygote expertise 🙂

PR #93, opened 2023-07-25
Update of #36 and #35

### LINKED ISSUE #34: Make value_and_pullback_function a primitive instead of pullback_function

### LINKED ISSUE #57: Code inside function with rrule should not run
MWE:

```
using Zygote
import ChainRulesCore.rrule
using ChainRulesCore: NoTangent
using AbstractDifferentiation

function myfunc(x)
    println("This should not print if I have an rrule.")
    x
end

rrule(::typeof(myfunc), x) = (x, (y -> (NoTangent(), y)))

println("Zygote run:")
Zygote.gradient(myfunc, 1) # nothing prints
println("AD run:")
AD.derivative(AD.ZygoteBackend(), myfunc, 1) # something prints
```
The code inside of myfunc should never run. In addition to possible inefficiency, this may lead to incorrect/confusing results for stateful functions or stochastic functions.

devmotion:
This PR is an update of #36, without merge conflicts and with updated primitives of Tracker and CRC.

A concrete example of why #36 and this PR are useful is #57: This issue is fixed if the primal is taken directly from `rrule` instead of manually re-computing the primal.

Fixes #57.

cc @sethaxen

Edit: I also updated and merged #35 in this PR to fix https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/93#discussion_r1279975890.

Fixes #34.


oxinabox (approved):
yes, this is much better

ext/AbstractDifferentiationChainRulesCoreExt.jl
```diff
@@ -3,11 +3,22 @@ module AbstractDifferentiationChainRulesCoreExt
...
 
-AD.@primitive function pullback_function(ba::AD.ReverseRuleConfigBackend, f, xs...)
-    _, back = ChainRulesCore.rrule_via_ad(AD.ruleconfig(ba), f, xs...)
-    pullback(vs) = Base.tail(back(vs))
-    pullback(vs::Tuple{Any}) = Base.tail(back(first(vs)))
-    return pullback
+AD.@primitive function value_and_pullback_function(ba::AD.ReverseRuleConfigBackend, f, xs...)
+    value, back = ChainRulesCore.rrule_via_ad(AD.ruleconfig(ba), f, xs...)
+    function value_and_pullback(vs)
+        pb_value = if vs === nothing
```
devmotion, line 9: We have to handle `nothing` due to the way how `jacobian` is implemented for `@primitive function value_and_pullback_function` (this PR) and `@primitive function pullback_function` (on the master branch): https://github.com/JuliaDiff/AbstractDifferentiation.jl/blob/041d760d4c0bff8a5e023f9d1d5cad2ffbb68f4d/src/AbstractDifferentiation.jl#L551

devmotion: On the master branch, the default fallback for `value_and_pullback_function` handles `nothing` inputs, hence the design choice becomes apparent only in this PR.

devmotion: I'm not sure why the `jacobian` is implemented in this way. In principle, calling `pullback` repeatedly should be sufficient and the `value` should not be needed.

devmotion: Could be solved easily with https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/35 I assume.

devmotion:
I solved https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/93#discussion_r1279975890 by merging and updating #35. With this change, `nothing` is not "abused" as co-tangent anymore in AbstractDifferentiation,  the code could be simplified quite a bit, and implementations of `value_and_pullback_function` do not have to handle `nothing` anymore.

gdalle:
@devmotion would it make sense to integrate #100 into the current PR?
There I tried to handle the tuple and the single-input cases correctly (to resolve #99), but I'm not confident enough with the package to decide whether I did it right. Also it needs testing but the AbstractDifferentiation.jl test base is still scary to me ^^

devmotion:
I guess it could but I'm hesitant right now because #100 still seems to be quite preliminary and #100 is supposed to be a bugfix that could be released in a non-breaking release whereas this PR is clearly breaking.

gdalle:
Then would you mind maybe reviewing it since it's rather short, and advising me on how to add tests? Since you've already worked on that part of the code, your opinion would be very valuable!

devmotion:
Bump :slightly_smiling_face: 

#95 could be included in a breaking release as well. And of course, I'd be fine with merging some of the other non-breaking PRs first if they are reviewed and approved :slightly_smiling_face:

src/AbstractDifferentiation.jl
```diff
@@ -221,48 +221,27 @@ end
...
-            vs = f(xs...)
-            if !primalcalled
-                value = primal_value(lowest(ab), vs, f, xs)
-                primalcalled = true
-            end
-            return value, nothing
-        end
-        pb = pullback_function(lowest(ab), (_xs...,) -> begin
+    value = f(xs...)
+    pbf = ws -> begin
```
gdalle, line 233: I think this would be clearer with actual named closures instead of anonymous functions. Plus that way they could use dispatch
```julia
pbf(ws::Tuple) =
pbf(ws) =
```
and make debugging easier for users

src/AbstractDifferentiation.jl
```diff
@@ -221,48 +221,27 @@ end
...
+        return gradient(lowest(ab), (_xs...,) -> begin
             vs = f(_xs...)
-            if !primalcalled
-                value = primal_value(lowest(ab), vs, f, xs)
-                primalcalled = true
+            if ws isa Tuple
+                @assert length(vs) == length(ws)
```
gdalle, line 237: See #102, this is buggy

devmotion: I know :slightly_smiling_face:  My impression, however, was that it will take a bit more time to finish #102 and I don't think it is necessary to hold back this PR due to it. IMO there are too many open PRs right now, so it would be good to merge every PR that is done - even if it does not fix every currently existing bug (my general feeling is that PRs in this repo are not merged timely enough, as evident by missing reviews and #35 and #36 having become outdated).

src/AbstractDifferentiation.jl
devmotion: I'm fine with using named functions but I think splitting up the function at the argument level here would lead to quite a bit of code duplication.

src/AbstractDifferentiation.jl
gdalle: That makes sense. Just wanted to point it out in case @mohamed82008 wanted to review, since he's also working on the fix!

devmotion: I would be very much in favour of this more drastic change. I haven't been involved in the initial design of the package, so possibly there are very good reasons for the current design. But I've always found it confusing - also as a user - that the design contains parts such as "If f has a single input, pf_f can also accept a single input instead of a 1-tuple". In my experience, such conventions tend to become painful for both users and developers - and often lead to ambiguity issues in special cases.

gdalle: https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/103

oxinabox:
Please do

devmotion:
I thought part of ColPrac is that developers merge their own PRs after approval? 🤔😛

gdalle:
I was planning to read it this morning! And actually your PRs had been waiting for so long I thought you didn't have merge rights, and I didn't even check 😅 sorry about that

gdalle:
Speaking of, does the repo have the right branch protections to help enforce ColPrac?

devmotion:
Well, they weren't approved (and I also wanted to wait for Mohammeds approval specifically). I merged and released #97 as soon as it was approved yesterday 🙂

gdalle:
I apologize for the confusion. I'm now up to speed on ColPrac ;)

PR #94, opened 2023-07-28
Remove special cases for `AbstractFiniteDifference`

devmotion:
Based on the discussion in #2, it seems the special cases for `AbstractFiniteDifference` only exist because FiniteDifferences mutated the inputs in previous releases (also, the checks don't seem completely consistent). This FiniteDifference behaviour was a bug and fixed in version 0.12.28 (https://github.com/JuliaDiff/FiniteDifferences.jl/pull/223).

This PR removes the special cases and checks if tests pass with the latest FiniteDifferences release.


devmotion:
Completely missed that of course in finite differencing the differentiated function might be called the first time with different inputs - and that this is the reason for the FiniteDifference branches, not the bug in FiniteDifferences.

Issue #91, opened 2023-07-25
Add docs for how to implement API as an AD


oxinabox:
There as no docs on `@primative` that i can find.
Nor any guidance anywhere else

devmotion:
Related: https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/13

gdalle:
It would also make sense to have a DAG of which functions depend on which others. If only because the roots are the minimal set of implementations that a backend must provide

PR #95, opened 2023-07-28
Remove the jacobian and primal_value primitives

devmotion:
The jacobian and the primal_value support for `@primitive` does not provide any benefit but IMO rather makes the code less readable (related also to https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/91): The macro version is equivalent to directly defining `AD.jacobian` or `AD.primal_value`.

Hence this PR proposes to remove support for `@primitive function jacobian` and `@primitive function primal_value` (which seems unused and untested).

### DIFF
--- a/ext/AbstractDifferentiationFiniteDifferencesExt.jl
+++ b/ext/AbstractDifferentiationFiniteDifferencesExt.jl
@@ -15,7 +15,7 @@ Create an AD backend that uses forward mode with FiniteDifferences.jl.
 """
 AD.FiniteDifferencesBackend() = AD.FiniteDifferencesBackend(FiniteDifferences.central_fdm(5, 1))
 
-AD.@primitive function jacobian(ba::AD.FiniteDifferencesBackend, f, xs...)
+function AD.jacobian(ba::AD.FiniteDifferencesBackend, f, xs...)
     return FiniteDifferences.jacobian(ba.method, f, xs...)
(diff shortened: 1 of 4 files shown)


oxinabox:
we should do this.
I guess it is technically breaking but since noone is using them, probably fine

mohamed82008:
I am in favour of breaking changes in this package and tagging a new release. I apologise for not being too actively involved here. Please tag team on this package without me and freely merge PRs if you all agree on the changes. I don't want to get in the way of progress here. Good luck!

devmotion:
Could someone officially approve the PR? 🙂

mohamed82008 (approved):

devmotion:
Thank you @mohamed82008! Makes me wonder though if we should adopt the colprac guidelines if more people are maintaining the package now (so far my impression was that basically everything should be approved by you 🙂)?

mohamed82008:
As the initiator of this package, I played the role of its maintainer for a while and then I got busy with other stuff and the package was not actively maintained. I think this package should be community-maintained since there seems to be enough interest and no time on my part to be the maintainer. I am not familiar with colprac guidelines for multiple maintainers. But I would say as long as 1 qualified reviewer approves the PR and no one objects in a certain time window, it can be safely merged. Happy to adopt to any other standard folks are used to though.

PR #96, opened 2023-07-29
Simpler defaults without FiniteDifferences special cases

devmotion:
IMO the AbstractFiniteDifference special cases in the default definitions are confusing (see #94...) and lead to code that is less idiomatic and more difficult to optimize.

With increasing dimensionality, avoiding one additional computation of the primal matters less and less, and hence this PR proposes to disentangle the computation of the primal and the jacobian/Hessian/etc. Moreover, also with these changes one can improve performance for individual backends, if possible and desired, by defining a more optimized `value_and_jacobian` etc.

An additional advantage for higher-order calls `value_and_hessian` and `value_gradient_and_hessian` is that it is sufficient to call `gradient` instead of `value_and_gradient`.

### DIFF
--- a/src/AbstractDifferentiation.jl
+++ b/src/AbstractDifferentiation.jl
@@ -66,21 +66,8 @@ function value_and_gradient(ab::AbstractBackend, f, xs...)
     return value, reshape.(adjoint.(jacs),size.(xs))
 end
 function value_and_jacobian(ab::AbstractBackend, f, xs...)
-    local value
-    primalcalled = false
-    if lowest(ab) isa AbstractFiniteDifference
-        value = primal_value(ab, nothing, f, xs)
-        primalcalled = true
-    end
-    jacs = jacobian(lowest(ab), (_xs...,) -> begin
-        v = f(_xs...)
-        if !primalcalled
-            value = primal_value(ab, v, f, xs)
-            primalcalled = true
-        end
-        return v
-    end, xs...)
-
+    value = f(xs...)
+    jacs = jacobian(lowest(ab), f, xs...)
     return value, jacs
 end
 function value_and_hessian(ab::AbstractBackend, f, x)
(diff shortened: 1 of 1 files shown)


gdalle:
I'm strongly in favor of more simplicity at the expense of a little performance.
I think there might be a conflict with #93 though, gonna try to do the merge

devmotion:
Ah, missed that there's a merge conflict. I'll fix it.

devmotion:
The merge conflicts are fixed, the PR is ready for review 🙂

src/AbstractDifferentiation.jl
```diff
@@ -480,8 +440,6 @@ macro primitive(expr)
         return define_jacobian_and_friends(fdef) |> esc
     elseif name == :primal_value
         return define_primal_value(fdef) |> esc
-    elseif name == :pullback_function
-        return define_pullback_function_and_friends(fdef) |> esc
```
gdalle, line 484: Why delete this?

gdalle (commented):
Looks good to me, apart from the deletion in the `@primitive` macro, which I don't understand.
Although to be fair I don't understand the macro as a whole. Why do we only have 2 out of 5 (now 4) `..._and_friends` functions defined in the package? Where are the rest?

src/AbstractDifferentiation.jl
devmotion: Because it's not supported anymore. It seems it was just forgotten to remove it in #93. I noticed it when fixing the merge conflicts. I could move it to a separate PR but I thought maybe it's fine to include this fix here.

devmotion: >  Why do we only have 2 out of 5 (now 4) ..._and_friends functions defined in the package? Where are the rest?

Good catch, the jacobian stuff is also missing. I'm not sure if the `jacobian_and_friends` has ever existed.

gdalle: So in the spirit of "simpler defaults", we might as well delete the rest?

devmotion: Ah, now I remember. The support for the jacobian macro was removed in https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/95. It seems it was reintroduced accidentally when fixing some merge conflicts?!

gdalle: So in the end, for a backend library to support AD, they either need to implement `pushforward_function` or `value_and_pullback_function`, from which the rest is deduced?
Making a mental note to document that and solve #91

devmotion: I guess you can use the macro and everything else will work. But I think usually that's not what you want. As already clear from the - probably incomplete - implementations for AD backends in this package and e.g. #97, good performance is only achieved if you implement all functions in the AbstractDifferentation interface for which the backend provides a - typically optimized - implementation.

The sole purpose of the macro is that you get an implementation of `jacobian` for free. But if the AD backend already defines a `jacobian` function, you can directly call it and don't even need that.

gdalle: Right, thanks! We can continue that conversation over on #91, in the meantime this is good to go

gdalle (approved):

PR #100, opened 2023-08-05
Fix tuple behavior in pushforward_function and pullback_function

gdalle:
Fix #99 

Needs tests to ensure non-regression

Probably useless after #93 but could be merged into it

Tried to make the code a bit more readable by avoiding chained anonymous functions

Issue #53, opened 2022-02-09
Use multiple arguments instead of a tuple for pushforward and pullback function?


devmotion:
It seems annoying that the pushforward and pullback function accept tuples of co-tangents instead of multiple arguments. Is there a compelling reason for doing so or was this a design decision that could be changed? In my opinion the main annoyance is that one has to handle the case of tuples of length 1 in a special way (as e.g. in https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/51) (it also makes it impossible to work with actual single-argument functions that take a tuple as only argument but maybe this is not needed anyway). Arguably it is also cleaner to provide multiple arguments as, well, multiple arguments instead of a tuple.

mohamed82008:
Yes I think this can be considered along with #35.

sethaxen:
> It seems annoying that the pushforward and pullback function accept tuples of co-tangents instead of multiple arguments. Is there a compelling reason for doing so or was this a design decision that could be changed?

While Julia functions may take multiple inputs, no Julia function returns multiple outputs. Instead, they might return a tuple of outputs. The (co)tangent of a tuple is like a tuple itself. FWIW, this is consistent with how ChainRules behaves, hence why in ChainRules, it would be represented as a `Tangent`, and here it would be represented as a tuple. One could make the case that since AD.jl supports only functions whose inputs and outputs are arrays, then if such a function returns a tuple it can only be interpreted as multiple outputs, but that would be inconsistent at least with ChainRules and Zygote.

> it also makes it impossible to work with actual single-argument functions that take a tuple as only argument but maybe this is not needed anyway

I don't think function with tuple inputs would be supported anyways.

devmotion:
> While Julia functions may take multiple inputs, no Julia function returns multiple outputs. Instead, they might return a tuple of outputs. The (co)tangent of a tuple is like a tuple itself. FWIW, this is consistent with how ChainRules behaves, hence why in ChainRules, it would be represented as a Tangent, and here it would be represented as a tuple. One could make the case that since AD.jl supports only functions whose inputs and outputs are arrays, then if such a function returns a tuple it can only be interpreted as multiple outputs, but that would be inconsistent at least with ChainRules and Zygote.

Sure, multiple outputs are in fact just a tuple of outputs. But it does not necessarily mean that we have to use a tuple as input to the pullback and pushforward function.

[...]

PR #102, opened 2023-08-06
Fix #99

### LINKED ISSUE #99: Outputs of `pushforward_function` and `pullback_function` mistake a single vector output for a tuple?

mohamed82008:
This PR fixes #99 and supersedes #100.


gdalle:
I think this will be superseded by #53 if we can manage it

PR #106, opened 2023-09-22
Move README content to docstrings & update docs website

gdalle:
README content is not discoverable by users in the REPL.
This PR fixes that by moving the function descriptions to docstrings, and updating the documentation website to reflect that.

- [x] move most README content to docstrings in the main source file
- [x] add links to backend docstrings 
- [x] update Documenter to v1.0 
- [x] ignore `docs/Manifest.toml` and activate CompatHelper on `docs/Project.toml`
- [x] activate PR docs preview: see <https://juliadiff.org/AbstractDifferentiation.jl/previews/PR106>
- [x] clean up resulting HTML once PR is closed
- [x] clarify dependencies between functions
- [x] explain what `@primitive` does and how to use it

No actual Julia code has been changed except in `docs/make.jl`


docs/make.jl
```diff
@@ -26,21 +27,23 @@ end
...
-        "API reference" => "api.md",
+        "User guide" => "user_guide.md",
+        "Implementer guide" => "implementer_guide.md",
     ],
 )
 
 deploydocs(;
     repo="github.com/JuliaDiff/AbstractDifferentiation.jl",
     devbranch="master",
+    push_preview=true,
```
devmotion, line 48: Could you add the Github action that removes these previews automatically?

devmotion (commented):
Great!

I think usually it's easier to not commit the Manifest file.

gdalle:
> I think this might require more discussion and changes in the code, so I'd suggest leaving it for a separate PR.

I agree, but still I think leaving the page in the docs makes sense, if only to show readers that there are two aspects to the package and only one is documented yet.

docs/make.jl
gdalle: do you have a specific one in mind?

devmotion: There's one in the Documenter docs that should work I assume (I guess over time possibly I accumulated smaller modifications in my repos but they should not matter here).

devmotion: https://documenter.juliadocs.org/stable/man/hosting/#gh-pages-Branch

src/AbstractDifferentiation.jl
```diff
@@ -7,19 +7,34 @@ abstract type AbstractFiniteDifference <: AbstractBackend end
...
+"""
+    HigherOrderBackend{B}
+
+Let `ab_f` be a forward-mode automatic differentiation backend and let `ab_r` be a reverse-mode automatic differentiation backend.
+To construct a higher order backend for doing forward-over-reverse-mode automatic differentiation, use `AD.HigherOrderBackend((ab_f, ab_r))`.
+To construct a higher order backend for doing reverse-over-forward-mode automatic differentiation, use `AD.HigherOrderBackend((ab_r, ab_f))`.
+
+# Fields
+
+- `backends::B`
```
devmotion, line 19: This could be automated with DocStringExtensions.

src/AbstractDifferentiation.jl
```diff
@@ -30,6 +45,13 @@ primal_value(::AbstractBackend, ys, ::Any, ::Any) = primal_value(ys)
 primal_value(x::Tuple) = map(primal_value, x)
 primal_value(x) = x
 
+"""
+    AD.derivative(ab::AD.AbstractBackend, f, xs::Number...)
+
+Compute the derivatives `ds` of `f` wrt the numbers `xs` using the backend `ab`. 
+
+Return a tuple `ds` of derivatives, one for each element in `xs`.
```
devmotion, line 53: The name is not needed and does not add any important information, I think? IMO (matter of taste) it's also a bit strange to use the imperative form twice.
```suggestion
Compute the derivatives of `f` with respect to the numbers `xs` using the backend `ab`.

The function returns a `Tuple` of derivatives, one for each element in `xs`.
```

src/AbstractDifferentiation.jl
```diff
@@ -39,14 +61,37 @@ function derivative(ab::AbstractBackend, f, xs::Number...)
     end
 end
 
+"""
+    AD.gradient(ab::AD.AbstractBackend, f, xs...)
+
+Compute the gradients `gs` of `f` wrt the inputs `xs` using the backend `ab`.
+
+Return a tuple `gs` of gradients, one for each element in `xs`.
```
devmotion, line 69:
```suggestion
Compute the gradients of `f` with respect to the inputs `xs` using the backend `ab`.

The function returns a `Tuple` of gradients, one for each element in `xs`.
```

src/AbstractDifferentiation.jl
```diff
@@ -39,14 +61,37 @@ function derivative(ab::AbstractBackend, f, xs::Number...)
...
 function gradient(ab::AbstractBackend, f, xs...)
     return reshape.(adjoint.(jacobian(lowest(ab), f, xs...)),size.(xs))
 end
+
+"""
+    AD.jacobian(ab::AD.AbstractBackend, f, xs...)
+
+Compute the Jacobians `js` of `f` wrt the inputs `xs` using the backend `ab`.
+
+Return a tuple `js` of Jacobians, one for each element in `xs`.
```
devmotion, line 80:
```suggestion
Compute the Jacobians of `f` with respect to the inputs `xs` using the backend `ab`.

The function returns a `Tuple` of Jacobians, one for each element in `xs`.
```

src/AbstractDifferentiation.jl
```diff
@@ -39,14 +61,37 @@ function derivative(ab::AbstractBackend, f, xs::Number...)
...
 function jacobian(ab::HigherOrderBackend, f, xs...) 
     jacobian(lowest(ab), f, xs...)
 end
 
+"""
+    AD.hessian(ab::AD.AbstractBackend, f, x)
+
+Compute the Hessian `h` of `f` wrt the input `x` using the backend `ab`.
+
+Return a single matrix `h`, because `hessian` currently only supports a single input.
```
devmotion, line 93:
```suggestion
Compute the Hessian of `f` wrt the input `x` using the backend `ab`.

The function returns a single matrix because `hessian` currently only supports a single input.
```

src/AbstractDifferentiation.jl
```diff
@@ -57,19 +102,51 @@ function hessian(ab::AbstractBackend, f, x)
     end, x)
 end
 
+
+"""
+    AD.value_and_derivative(ab::AD.AbstractBackend, f, xs::Number...)
+
+Compute the function value `v = f(xs...)` and the derivatives `ds` of `f` wrt the numbers `xs` using the backend `ab`.
+    
+Return a tuple `(v, ds)` where `ds` is a tuple of derivatives, one for each element in `xs`.
```
devmotion, line 111: I think this is maybe a bit repetetive, which also makes it a bit less convenient when making changes to e.g. `derivative. Maybe mainly refer to the other docstring?
```suggestion
Return the tuple `(v, ds)` of the function value `v = f(xs...)` and the derivatives `ds = AD.derivative(ab, f, xs...).

See also [`AD.derivative`](@ref).
```

src/AbstractDifferentiation.jl
```diff
@@ -57,19 +102,51 @@ function hessian(ab::AbstractBackend, f, x)
...
     value, jacs = value_and_jacobian(lowest(ab), f, xs...)
     return value[1], getindex.(jacs, 1)
 end
+
+"""
+    AD.value_and_gradient(ab::AD.AbstractBackend, f, xs...)
+
+Compute the function value `v = f(xs...)` and the gradients `gs` of `f` wrt the inputs `xs` using the backend `ab`.
+    
+Return a tuple `(v, gs)` where `gs` is a tuple of gradients, one for each element in `xs`.
```
devmotion, line 123:
```suggestion
Return the tuple `(v, gs)` of the function value `v = f(xs...)` and the gradients `gs = AD.gradient(ab, f, xs...).
    
See also [`AD.gradient`](@ref).
```

src/AbstractDifferentiation.jl
```diff
@@ -57,19 +102,51 @@ function hessian(ab::AbstractBackend, f, x)
...
     value, jacs = value_and_jacobian(lowest(ab), f, xs...)
     return value, reshape.(adjoint.(jacs),size.(xs))
 end
+
+"""
+    AD.value_and_jacobian(ab::AD.AbstractBackend, f, xs...)
+
+Compute the function value `v = f(xs...)` and the Jacobians `js` of `f` wrt the inputs `xs` using the backend `ab`.
+    
+Return a tuple `(v, js)` where `js` is a tuple of Jacobians, one for each element in `xs`.
```
devmotion, line 135:
```suggestion
Return the tuple `(v, Js)` of the function value `v = f(xs...)` and the Jacobians `Js = AD.jacobian(ab, f, xs...)`.
    
See also [`AD.jacobian`](@ref).
```

src/AbstractDifferentiation.jl
```diff
@@ -57,19 +102,51 @@ function hessian(ab::AbstractBackend, f, x)
...
     jacs = jacobian(lowest(ab), f, xs...)
     return value, jacs
 end
+
+"""
+    AD.value_and_hessian(ab::AD.AbstractBackend, f, x)
+
+Compute the function value `v = f(x)` and the Hessian `h` of `f` wrt the input `x` using the backend `ab`.
+    
+Return a tuple `(v, h)` where `h` is a single matrix, because `hessian` currently only supports a single input.
```
devmotion, line 148:
```suggestion
Return the tuple `(v, H)` of the function value `v = f(x)` and the Hessian `H = AD.hessian(ab, f, x)`.

See also [`AD.hessian`](@ref).    
```

src/AbstractDifferentiation.jl
```diff
@@ -98,6 +176,14 @@ function value_and_hessian(ab::HigherOrderBackend, f, x)
 
     return value, hess
 end
+
+"""
+    AD.value_gradient_and_hessian(ab::AD.AbstractBackend, f, x)
+    
+Compute the function value `v = f(x)` and the gradient `g` and Hessian `h` of `f` wrt the input `x` using the backend `ab`.
+    
+Return a tuple `(v, g, h)` where `g` is a single vector and `h` is a single matrix, because `hessian` currently only supports a single input.
```
devmotion, line 185:
```suggestion
Return the tuple `(v, g, H)` of the function value `v = f(x)`, the gradient `g = AD.gradient(ab, f, x)`, and the Hessian `H = AD.hessian(ab, f, x)`.

See also [`AD.gradient`](@ref) and [`AD.hessian`](@ref).
```

src/AbstractDifferentiation.jl
```diff
@@ -127,6 +214,14 @@ function value_gradient_and_hessian(ab::HigherOrderBackend, f, x)
     return value, (grads,), hess
 end
 
+"""
+    AD.pushforward_function(ab::AD.AbstractBackend, f, xs...)
+    
+Returns the pushforward function `pf_f` of the function `f` at the inputs `xs`. 
+    
+`pf_f` is a function that accepts the tangents `vs` as input which is a tuple of length equal to the length of the tuple `xs`.
+If `f` has a single input, `pf_f` can also accept a single input instead of a 1-tuple.
```
devmotion, line 223: I think it reads a bit easier without underscores (and also for gradients, Jacobians, hessians, etc. we don't explicitly mention `f` in the notation).
```suggestion
Return the pushforward function `pf` of the function `f` at the inputs `xs` using backend `ab`. 
    
The pushfoward function `pf` accepts as input a `Tuple` of tangents, one for each element in `xs`.
If `xs` consists of a single element, `pf` can also accept a single tangent instead of a 1-tuple.
```

src/AbstractDifferentiation.jl
```diff
@@ -145,6 +240,16 @@ function pushforward_function(
...
 end
+
+"""
+    AD.value_and_pushforward_function(ab::AD.AbstractBackend, f, xs...)
+    
+Return a function `value_and_pf_f` that computes a 2-tuple.
+This tuple contains the value `f(xs...)` and the output of the pushforward function `pf_f`.
+    
+`value_and_pf_f` is a function that accepts the tangent `vs` as input, which is a tuple of length equal to the length of the tuple `xs`.
+If `f` has a single input, `value_and_pf_f` can accept a single input instead of a 1-tuple.
```
devmotion, line 251:
```suggestion
Return a function that computes the tuple `(v, p)` of the function value `v = f(xs...)` and the output `p` of the pushforward function `AD.pushforward_function(ab, f, xs...)`.

See also [`AD.pushforward_function`](@ref).
```

devmotion (commented):
I have to turn to something else for a while but I had a look at large parts of the PR. Here are some initial comments (that can probably be extrapolated).

docs/Project.toml
```diff
@@ -1,3 +1,4 @@
 [deps]
 AbstractDifferentiation = "c29ec348-61ec-40c8-8164-b8c60e9d9f3d"
```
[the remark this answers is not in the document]
gdalle: I didn't know that, do you have a source? I don't see how the docs build can work if the original package is not a dependency. I `>pkg dev` the package into all docs envs

docs/make.jl
```diff
@@ -1,7 +1,8 @@
 using AbstractDifferentiation
```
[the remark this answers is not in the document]
gdalle: The name `AbstractDifferentiation` and the abbreviation `AD` are both used in various spots

docs/make.jl
```diff
@@ -26,21 +27,23 @@ end
 makedocs(;
     modules=[AbstractDifferentiation],
     authors="Mohamed Tarek <mohamed82008@gmail.com> and contributors",
-    repo="https://github.com/JuliaDiff/AbstractDifferentiation.jl/blob/{commit}{path}#{line}",
     sitename="AbstractDifferentiation.jl",
     format=Documenter.HTML(;
+    repolink="https://github.com/JuliaDiff/AbstractDifferentiation.jl",
```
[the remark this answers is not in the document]
gdalle: That's the point of updating to Documenter v1.0, strict checks are enabled by default now

docs/Project.toml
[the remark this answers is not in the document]
devmotion: Yes, but that's done automatically if you use the standard Github action for building docs.

src/AbstractDifferentiation.jl
gdalle: I would do that in a separate PR
Opening https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/107 to keep track

src/AbstractDifferentiation.jl
gdalle: I tried to copy-paste as much as possible from the README, leaving style improvements to future PRs once the documentation is in place (diffs will be easier to analyze). What do you think?

docs/Project.toml
```diff
@@ -1,3 +1,4 @@
...
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
```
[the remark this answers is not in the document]
gdalle: The question I always ask myself is: when we don't include `docs/Manifest.toml`, how is Julia supposed to know we want to `dev` the current code of the package into the docs environment instead of downloading the stable version from the general registry?

src/AbstractDifferentiation.jl
devmotion: But there are no docstrings yet, are there? I thought this is a new addition.

gdalle: Since you already reviewed the style, I might as well commit your changes. Just explaining that my goal was only moving stuff around at first ;)

docs/Project.toml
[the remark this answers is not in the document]
gdalle: Right but when I build the docs locally on my computer it needs a correct Manifest to start

src/AbstractDifferentiation.jl
gdalle: It is a new feature which copypastes exactly what was in the README. My thought process was moving stuff first, improving stuff next. But we're on track to do both so let's forge ahead!

docs/Project.toml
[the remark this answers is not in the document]
devmotion: Well, you specify it 😅 That's exactly what julia-docdeploy (https://github.com/julia-actions/julia-docdeploy/blob/580a86b97cef20258d96b53292326e6258dd6859/action.yml#L30) and basically every custom Julia doc build does.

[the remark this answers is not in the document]
gdalle: for CI I understand, but locally that assumes the developer already has a working manifest, otherwise when they activate and instantiate the docs environment that will by default download the stable version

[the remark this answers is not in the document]
devmotion: Typically, locally you also first `dev` the package. I have really bad experiences with committing Manifest files for docs, so I'm strongly convinced that we should save ourselves the headaches that come with it and only add a Project.toml file.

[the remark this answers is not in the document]
gdalle: alright, I trust you
gonna look up how to add CompatHelper for a subfolder

gdalle:
thanks for the huge merge conflict from JuliaFormatter 🤣

docs/Project.toml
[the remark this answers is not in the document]
devmotion: https://juliaregistries.github.io/CompatHelper.jl/dev/options/#Packages-in-subdirectories

devmotion:
Ooops, sorry 😬

docs/Project.toml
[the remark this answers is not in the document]
gdalle: Actually when you instantiate the docs environment locally and then dev the package into it, this line will appear in the Project anyway. Why is it a good idea to remove it?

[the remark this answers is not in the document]
gdalle: I understand your point about the Manifest but I think this modification of the Project only works when you never build the docs locally

gdalle:
I encountered a weird bug where the cross-references in the docs wouldn't work with the `AD` shortcut for `AbstractDifferentiation`, even though I imported it as such in `docs/make.jl`. This explains the failure of the workflows on the penultimate commit. Do you think it's a Documenter bug?

devmotion (commented):
> I encountered a weird behavior where the cross-references in the docs wouldn't work with the AD shortcut for AbstractDifferentiation, even though I imported it as such in docs/make.jl. This explains the workflow failure on the penultimate commit. Do you think it's a Documenter bug or just me asking too much?

Based on the CI log, Documenter tries to resolve the references within the `AbstractDifferentiation`/`AD` module - but of course `AD` does not exist there, only `AbstractDifferentiation`. So maybe one option would be to set the current module to `Main`.

Generally, I think it's not too bad (maybe even clearer?) if the docs do only use the user-defined convenience alias in the examples. On the other hand, it creates a slight inconsistency...

.gitignore
```diff
@@ -1,3 +1,4 @@
 /Manifest.toml
 /docs/build
-/docs/src/index.md
\ No newline at end of file
+/docs/src/index.md
+/docs/Manifest.toml
```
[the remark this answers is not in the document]
gdalle: we might need other manifests later on, e.g. for fully reproducible benchmarks

gdalle:
> Generally, I think it's not too bad (maybe even clearer?) if the docs do only use the user-defined convenience alias in the examples.

I do agree, but I don't think it should block this PR.

> So maybe one option would be to set the current module to Main.

I looked around and I didn't find a way to do that in the kwargs of `Documenter.makedocs`, are you aware of one?
I would proceed with caution though, cause it might trigger other unexpected behaviors. Maybe it is actually easier to define `const AD = AbstractDifferentiation` within the module itself, without exporting it?

devmotion:
>  Maybe it is actually easier to define const AD = AbstractDifferentiation within the module itself, without exporting it?

This was done initially but caused some problems (would have to look it up though). Generally, my gut feeling is that it is cleaner and easier to let users decide whether and what abbreviation they want to use. `import ... as ...` was designed for that use case and seems sufficiently concise IMO.

devmotion (approved):

gdalle:
Thanks for the thorough review!

PR #108, opened 2023-09-22
Add JuliaFormatter

devmotion:
I'm a big fan of JuliaFormatter, in particular when working together with others. It makes it just easier to focus on the content instead of having to deal with formatting preferences and changes.

In this PR I set up JuliaFormatter and a Github action that makes formatting suggestions in PRs (used e.g., in ChainRules, JuliaGaussianProcesses, some Turing packages (IIRC), and some of my packages). I went for Blue style because I use that often but I'm also fine with other styles (just not a fan of the long lines and strange indentation in the SciML format 😛).

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,3 @@
+style = "blue"
+format_markdown = true
+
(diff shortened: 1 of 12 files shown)


gdalle (approved):
I've wanted to do it myself for a while ;)

PR #115, opened 2023-10-05
Fix docs

devmotion:
The PR tries to fix the build failure on the master branch (see https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/112#issuecomment-1749687780).

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -4,6 +4,6 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [compat]
-AbstractDifferentiation = "0.5"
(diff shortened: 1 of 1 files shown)


gdalle:
I think we should just keep 0.6 since the docs will no longer be accurate for 0.5? Or remove the project entry altogether as you suggested, honestly it's fine either way. It's too small an issue to waste time debating it

devmotion:
Just removing it won't play nicely with CompatHelper (and without any compat entries at all docs might break at any time, as happened in multiple repos with Documenter 1). So I guess the best is to just use 0.6.

devmotion:
Ah wait, that also won't work with XompatHelper - it will reopen PRs for 0.5. So 0.5, 0.6 is the best option I think.

gdalle (approved):

PR #117, opened 2023-10-12
Bump version to 0.6

gdalle:


devmotion (approved):
Maybe use the opportunity to update the compat entry in docs/Project.toml?

devmotion:
Yes, exactly. Once the new version is registered, CompatHelper won't try to add compat with 0.5 anymore.

devmotion (approved):

PR #118, opened 2023-10-12
Run CI on tags

gdalle:
Hopefully fixes the lack of a stable docs build

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -29,6 +29,7 @@ AbstractDifferentiationZygoteExt = "Zygote"
 [compat]
 ChainRulesCore = "1"
 DiffResults = "1"
+Documenter = "1"
 ExprTools = "0.1"
 FiniteDifferences = "0.12"
 ForwardDiff = "0.10"


devmotion:
Why don't we have a separate Docs.yml action? There's no need to rerun CI on tags, we only need the docs.

gdalle:
I guess it doesn't hurt to prove that the tests pass specifically for the tag you want. Otherwise if you wanna check you have to go back in time and look for the tests of the associated commit. But I don't feel strongly about either option

devmotion:
I think it's sufficient to run CI once for every commit. If you go to a tag, you also end up with the commit with the CI runs next to it.

gdalle:
Alright, I'll split it a docs workflow tomorrow

.github/workflows/Docs.yml
```diff
@@ -0,0 +1,42 @@
...
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: '1'
+      - name: Configure doc environment
+        run: |
+          julia --project=docs/ -e '
+            using Pkg
+            Pkg.develop(PackageSpec(path=pwd()))
+            Pkg.instantiate()'
+      - uses: julia-actions/julia-buildpkg@v1
```
devmotion, line 33: This should just be
```suggestion
      - uses: julia-actions/julia-buildpkg@v1
```
The `Pkg.develop` etc. is already part of the `julia-docdeploy` action: https://github.com/julia-actions/julia-docdeploy/blob/580a86b97cef20258d96b53292326e6258dd6859/action.yml#L28-L56

.github/workflows/Docs.yml
```diff
@@ -0,0 +1,42 @@
...
+      - uses: julia-actions/julia-docdeploy@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - run: |
+          julia --project=docs -e '
+            using Documenter: DocMeta, doctest
+            using AbstractDifferentiation
+            DocMeta.setdocmeta!(AbstractDifferentiation, :DocTestSetup, :(using AbstractDifferentiation); recursive=true)
+            doctest(AbstractDifferentiation)'
```
devmotion, line 42: Shouldn't we just run doctests as part of the tests in runtests.jl?

gdalle: Good idea, just added that

Project.toml
```diff
@@ -41,6 +41,7 @@ julia = "1.6"
 [extras]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 DiffResults = "163ba53b-c6d8-5494-b064-1a9d43ac40c5"
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
```
devmotion, line 44: Maybe it would be safer to add a compat entry for Documenter to Project.toml as well?

devmotion (approved):
Looks good!

Project.toml
gdalle: I'm unsure about the role of compat entries for non-dependencies. Do they constrain package users who also happen to have Documenter in their environment, or just the version we use for testing AbstractDifferentiation?

PR #122, opened 2023-12-20
Add second-order derivative functions

### LINKED ISSUE #71: Missing `value_derivative_and_second_derivative`?
Is there a scalar equivalent of `value_gradient_and_hessian` for functions that accept a `Number` rather than an `AbstractArray`?

That seems an asymmetric omission from the interface.

gerlero:
Draft PR for changes discussed in https://github.com/JuliaDiff/ForwardDiff.jl/pull/678

Fixes #71.

### DIFF
--- a/docs/src/implementer_guide.md
+++ b/docs/src/implementer_guide.md
@@ -34,9 +34,9 @@ They are just listed here to help readers figure out the code structure:
   - `value_and_derivative` calls `value_and_jacobian`
   - `value_and_gradient` calls `value_and_jacobian`
   - `value_and_hessian` calls `jacobian` and `gradient`
-  - `value_and_second_derivative` calls `second_derivative`
+  - `value_and_second_derivative` calls `secondderivative`
   - `value_gradient_and_hessian` calls `value_and_jacobian` and `gradient`
-  - `value_derivative_and_second_derivative` calls `value_and_derivative` and `second_derivative`
+  - `value_derivative_and_second_derivative` calls `value_and_derivative` and `secondderivative`
   - `pushforward_function` calls `jacobian`
   - `value_and_pushforward_function` calls `pushforward_function`
   - `pullback_function` calls `value_and_pullback_function`
--- a/docs/src/implementer_guide.md
+++ b/docs/src/implementer_guide.md
@@ -29,14 +29,14 @@ They are just listed here to help readers figure out the code structure:
   - `derivative` calls `jacobian`
   - `gradient` calls `jacobian`
   - `hessian` calls `jacobian` and `gradient`
-  - `second_derivative` calls `derivative`
+  - `secondderivative` calls `derivative`
   - `value_and_jacobian` calls `jacobian`
   - `value_and_derivative` calls `value_and_jacobian`
   - `value_and_gradient` calls `value_and_jacobian`
   - `value_and_hessian` calls `jacobian` and `gradient`
   - `value_and_second_derivative` calls `second_derivative`
   - `value_gradient_and_hessian` calls `value_and_jacobian` and `gradient`
-  - `value_and_derivatives` calls `value_and_derivative` and `second_derivative`
+  - `value_derivative_and_second_derivative` calls `value_and_derivative` and `second_derivative`
   - `pushforward_function` calls `jacobian`
   - `value_and_pushforward_function` calls `pushforward_function`
   - `pullback_function` calls `value_and_pullback_function`
--- a/docs/src/implementer_guide.md
+++ b/docs/src/implementer_guide.md
@@ -29,11 +29,14 @@ They are just listed here to help readers figure out the code structure:
   - `derivative` calls `jacobian`
   - `gradient` calls `jacobian`
   - `hessian` calls `jacobian` and `gradient`
+  - `second_derivative` calls `derivative`
   - `value_and_jacobian` calls `jacobian`
   - `value_and_derivative` calls `value_and_jacobian`
   - `value_and_gradient` calls `value_and_jacobian`
   - `value_and_hessian` calls `jacobian` and `gradient`
+  - `value_and_second_derivative` calls `second_derivative`
   - `value_gradient_and_hessian` calls `value_and_jacobian` and `gradient`
+  - `value_and_derivatives` calls `value_and_derivative` and `second_derivative`
   - `pushforward_function` calls `jacobian`
   - `value_and_pushforward_function` calls `pushforward_function`
   - `pullback_function` calls `value_and_pullback_function`
--- a/docs/src/implementer_guide.md
+++ b/docs/src/implementer_guide.md
@@ -29,14 +29,14 @@ They are just listed here to help readers figure out the code structure:
   - `derivative` calls `jacobian`
   - `gradient` calls `jacobian`
   - `hessian` calls `jacobian` and `gradient`
-  - `secondderivative` calls `derivative`
+  - `second_derivative` calls `derivative`
   - `value_and_jacobian` calls `jacobian`
   - `value_and_derivative` calls `value_and_jacobian`
   - `value_and_gradient` calls `value_and_jacobian`
   - `value_and_hessian` calls `jacobian` and `gradient`
-  - `value_and_second_derivative` calls `secondderivative`
+  - `value_and_second_derivative` calls `second_derivative`
   - `value_gradient_and_hessian` calls `value_and_jacobian` and `gradient`
-  - `value_derivative_and_second_derivative` calls `value_and_derivative` and `secondderivative`
+  - `value_derivative_and_second_derivative` calls `value_and_derivative` and `second_derivative`
   - `pushforward_function` calls `jacobian`
   - `value_and_pushforward_function` calls `pushforward_function`
   - `pullback_function` calls `value_and_pullback_function`
--- a/ext/AbstractDifferentiationForwardDiffExt.jl
+++ b/ext/AbstractDifferentiationForwardDiffExt.jl
@@ -61,6 +61,12 @@ function AD.hessian(ba::AD.ForwardDiffBackend, f, x::AbstractArray)
     return (ForwardDiff.hessian(f, x, cfg),)
 end
 
+function AD.value_and_derivative(::AD.ForwardDiffBackend, f, x::Real)
+    T = typeof(ForwardDiff.Tag(f, typeof(x)))
+    ydual = f(ForwardDiff.Dual{T}(x, one(x)))
+    return ForwardDiff.value(T, ydual), (ForwardDiff.extract_derivative(T, ydual),)
+end
+
 function AD.value_and_gradient(ba::AD.ForwardDiffBackend, f, x::AbstractArray)
     result = DiffResults.GradientResult(x)
     cfg = ForwardDiff.GradientConfig(f, x, chunk(ba, x))
@@ -68,11 +74,25 @@ function AD.value_and_gradient(ba::AD.ForwardDiffBackend, f, x::AbstractArray)
     return DiffResults.value(result), (DiffResults.derivative(result),)
 end
 
+function AD.value_and_second_derivative(ba::AD.ForwardDiffBackend, f, x::Real)
+    T = typeof(ForwardDiff.Tag(f, typeof(x)))
+    ydual, ddual = AD.value_and_derivative(ba, f, ForwardDiff.Dual{T}(x, one(x)))
+    return ForwardDiff.value(T, ydual), (ForwardDiff.extract_derivative(T, ddual[1]),)
+end
+
 function AD.value_and_hessian(ba::AD.ForwardDiffBackend, f, x)
     result = DiffResults.HessianResult(x)
     cfg = ForwardDiff.HessianConfig(f, result, x, chunk(ba, x))
     ForwardDiff.hessian!(result, f, x, cfg)
     return DiffResults.value(result), (DiffResults.hessian(result),)
+end
+
+function AD.value_and_derivatives(ba::AD.ForwardDiffBackend, f, x::Real)
+    T = typeof(ForwardDiff.Tag(f, typeof(x)))
+    ydual, ddual = AD.value_and_derivative(ba, f, ForwardDiff.Dual{T}(x, one(x)))
+    return ForwardDiff.value(T, ydual),
+    (ForwardDiff.value(T, ddual[1]),),
+    (ForwardDiff.extract_derivative(T, ddual[1]),)
 end
 
 @inline step_toward(x::Number, v::Number, h) = x + h * v
--- a/ext/AbstractDifferentiationForwardDiffExt.jl
+++ b/ext/AbstractDifferentiationForwardDiffExt.jl
@@ -79,8 +79,9 @@ function AD.value_and_second_derivative(ba::AD.ForwardDiffBackend, f, x::Real)
     xdual = ForwardDiff.Dual{T}(x, one(x))
     T2 = typeof(ForwardDiff.Tag(f, typeof(xdual)))
     ydual = f(ForwardDiff.Dual{T2}(xdual, one(xdual)))
-    return ForwardDiff.value(T, ForwardDiff.value(T2, ydual)),
-    (ForwardDiff.partials(T, ForwardDiff.partials(T2, ydual, 1), 1),)
+    v = ForwardDiff.value(T, ForwardDiff.value(T2, ydual))
+    d2 = ForwardDiff.partials(T, ForwardDiff.partials(T2, ydual, 1), 1)
+    return v, (d2,)
 end
 
 function AD.value_and_hessian(ba::AD.ForwardDiffBackend, f, x)
@@ -95,9 +96,10 @@ function AD.value_derivative_and_second_derivative(ba::AD.ForwardDiffBackend, f,
     xdual = ForwardDiff.Dual{T}(x, one(x))
     T2 = typeof(ForwardDiff.Tag(f, typeof(xdual)))
     ydual = f(ForwardDiff.Dual{T2}(xdual, one(xdual)))
-    return ForwardDiff.value(T, ForwardDiff.value(T2, ydual)),
-    (ForwardDiff.partials(T, ForwardDiff.value(T2, ydual), 1),),
-    (ForwardDiff.partials(T, ForwardDiff.partials(T2, ydual, 1), 1),)
+    v = ForwardDiff.value(T, ForwardDiff.value(T2, ydual))
+    d = ForwardDiff.partials(T, ForwardDiff.value(T2, ydual), 1)
+    d2 = ForwardDiff.partials(T, ForwardDiff.partials(T2, ydual, 1), 1)
+    return v, (d,), (d2,)
 end
 
 @inline step_toward(x::Number, v::Number, h) = x + h * v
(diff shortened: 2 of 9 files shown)


gdalle:
I think the change is welcome here and coherent regardless of whether it makes it into ForwardDiff. However it probably needs tests too, which I guess is why the PR remains a draft.
Thank you so much for taking the time!

gerlero:
@gdalle Thanks. Tests are now live and passing so this is ready for review

gerlero:
Gentle bump @gdalle; can you be a reviewer in this PR?

gdalle (commented):
Thanks for the contribution!
The backend-agnostic part looks good to me. I would have to read up on ForwardDiff tags to understand what you did in the relevant extension, so if someone else can review that part great, if not I'll do it too.
Any reason why you didn't add tests for `Tracker` and `RuleConfig` backends?

ext/AbstractDifferentiationForwardDiffExt.jl
```diff
@@ -61,20 +61,40 @@ function AD.hessian(ba::AD.ForwardDiffBackend, f, x::AbstractArray)
...
     result = DiffResults.GradientResult(x)
     cfg = ForwardDiff.GradientConfig(f, x, chunk(ba, x))
     ForwardDiff.gradient!(result, f, x, cfg)
     return DiffResults.value(result), (DiffResults.derivative(result),)
 end
 
+function AD.value_and_second_derivative(ba::AD.ForwardDiffBackend, f, x::Real)
+    T = typeof(ForwardDiff.Tag(f, typeof(x)))
+    ydual, ddual = AD.value_and_derivative(ba, f, ForwardDiff.Dual{T}(x, one(x)))
+    return ForwardDiff.value(T, ydual), (ForwardDiff.extract_derivative(T, ddual[1]),)
```
devmotion, line 80: This will hit the less efficient fallback method since the implementation of `value_and_derivative` above is only defined for `x::Real` but not for `ForwardDiff.Dual`s. Instead you could use e.g.
```suggestion
    xdual = ForwardDiff.Dual{T}(x, one(x))
    T2 = typeof(ForwardDiff.Tag(f, typeof(xdual)))
    ydual = f(ForwardDiff.Dual{T2}(xdual, one(xdual)))
    return ForwardDiff.value(T, ForwardDiff.value(T2, ydual)), (ForwardDiff.partials(T, ForwardDiff.partials(T2, ydual, 1), 1),)
```

ext/AbstractDifferentiationForwardDiffExt.jl
```diff
@@ -61,20 +61,40 @@ function AD.hessian(ba::AD.ForwardDiffBackend, f, x::AbstractArray)
...
     ForwardDiff.hessian!(result, f, x, cfg)
     return DiffResults.value(result), (DiffResults.hessian(result),)
 end
 
+function AD.value_and_derivatives(ba::AD.ForwardDiffBackend, f, x::Real)
+    T = typeof(ForwardDiff.Tag(f, typeof(x)))
+    ydual, ddual = AD.value_and_derivative(ba, f, ForwardDiff.Dual{T}(x, one(x)))
+    return ForwardDiff.value(T, ydual),
+    (ForwardDiff.value(T, ddual[1]),),
+    (ForwardDiff.extract_derivative(T, ddual[1]),)
```
devmotion, line 95: Similar to above, this will hit the fallback definition. I'd suggest
```suggestion
    xdual = ForwardDiff.Dual{T}(x, one(x))
    T2 = typeof(ForwardDiff.Tag(f, typeof(xdual)))
    ydual = f(ForwardDiff.Dual{T2}(xdual, one(xdual)))
    return ForwardDiff.value(T, ForwardDiff.value(T2, ydual)),
    (ForwardDiff.partials(T, ForwardDiff.value(T2, ydual), 1),),
    (ForwardDiff.partials(T, ForwardDiff.partials(T2, ydual, 1), 1),)
```
or equivalently
```suggestion
    xdual = ForwardDiff.Dual{T}(x, one(x))
    T2 = typeof(ForwardDiff.Tag(f, typeof(xdual)))
    ydual = f(ForwardDiff.Dual{T2}(xdual, one(xdual)))
    return ForwardDiff.value(T, ForwardDiff.value(T2, ydual)),
    (ForwardDiff.value(T, ForwardDiff.partials(T2, ydual, 1)),),
    (ForwardDiff.partials(T, ForwardDiff.partials(T2, ydual, 1), 1),)
```

src/AbstractDifferentiation.jl
```diff
@@ -176,6 +205,15 @@ function value_and_hessian(ab::HigherOrderBackend, f, x)
     return value, hess
 end
 
+"""
+    AD.value_and_derivatives(ab::AD.AbstractBackend, f, x)
+
+Return the tuple `(v, d, d2)` of the function value `v = f(x)` and the first and second derivatives `d = AD.derivative(ab, f, x)` and `d2 = AD.second_derivative(ab, f, x)`.
+"""
+function value_and_derivatives(ab::AbstractBackend, f, x)
+    return value_and_derivative(ab, f, x)..., second_derivative(ab, f, x)[1]
```
devmotion, line 214: This fallback is inconsistent with `value_gradient_and_hessian` which computes the first and second derivative in one go. I think we should do the same here and use
```suggestion
    if x isa Tuple
        # only support computation of Hessian for functions with single input argument
        x = only(x)
    end

    value = f(x)
    deriv, second_deriv = value_and_derivative(
        second_lowest(ab), _x -> begin
            d = derivative(lowest(ab), f, _x)
            return d[1] # derivative returns a tuple
        end, x
    )

    return value, (deriv,), second_deriv
```

ext/AbstractDifferentiationForwardDiffExt.jl
```diff
@@ -61,20 +61,40 @@ function AD.hessian(ba::AD.ForwardDiffBackend, f, x::AbstractArray)
     return (ForwardDiff.hessian(f, x, cfg),)
 end
 
+function AD.value_and_derivative(::AD.ForwardDiffBackend, f, x::Real)
+    T = typeof(ForwardDiff.Tag(f, typeof(x)))
```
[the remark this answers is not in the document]
gerlero: Sounds like a useful improvement, but should we leave it for a different PR?

ext/AbstractDifferentiationForwardDiffExt.jl
```diff
@@ -61,20 +61,40 @@ function AD.hessian(ba::AD.ForwardDiffBackend, f, x::AbstractArray)
...
+    ydual = f(ForwardDiff.Dual{T}(x, one(x)))
```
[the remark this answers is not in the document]
gerlero: Is your concern here that `f` might return an `AbstractArray`? If so I guess we could add a typeassert that `ydual` is a `Number`...

gerlero:
@gdalle @devmotion Thanks for the reviews.

> Any reason why you didn't add tests for Tracker and RuleConfig backends?

I simply omitted them because I saw that those backends don't have tests for `hessian` either. I've now tried adding the tests: `RuleConfig` passes, but `Tracker` fails (`ArgumentError: Tracker backend does not support nested differentiation.`). So, I've now added the `RuleConfig` backend tests, but left `Tracker` out.

docs/src/implementer_guide.md
```diff
@@ -29,11 +29,14 @@ They are just listed here to help readers figure out the code structure:
...
   - `gradient` calls `jacobian`
   - `hessian` calls `jacobian` and `gradient`
+  - `second_derivative` calls `derivative`
   - `value_and_jacobian` calls `jacobian`
   - `value_and_derivative` calls `value_and_jacobian`
   - `value_and_gradient` calls `value_and_jacobian`
   - `value_and_hessian` calls `jacobian` and `gradient`
+  - `value_and_second_derivative` calls `second_derivative`
   - `value_gradient_and_hessian` calls `value_and_jacobian` and `gradient`
+  - `value_and_derivatives` calls `value_and_derivative` and `second_derivative`
```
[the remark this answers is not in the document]
gerlero: I've now left it as:

- `secondderivative`
- `value_and_second_derivative`
- `value_and_derivative_and_second_derivative`

But I don't really like the inconsistency. I don't have a preference between `second_derivative` vs. `secondderivative`; but if we're going with the later then we should rename the others as `value_and_secondderivative` and `value_and_derivative_and_secondderivative`?

ext/AbstractDifferentiationForwardDiffExt.jl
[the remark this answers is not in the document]
gdalle: tracked in https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/127, closing here

ext/AbstractDifferentiationForwardDiffExt.jl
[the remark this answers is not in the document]
gdalle: do we check that anywhere else?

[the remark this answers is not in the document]
gerlero: > do we check that anywhere else?

@gdalle Not that I could see (see e.g. the [definition of `derivative` in `ForwardDiff`](https://github.com/JuliaDiff/ForwardDiff.jl/blob/5e9240ae63e5d4217e1dc17cc14deca1d74c1ad7/src/derivative.jl#L12-L15))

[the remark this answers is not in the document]
gdalle: ok then let's not break the pattern

ext/AbstractDifferentiationForwardDiffExt.jl
```diff
@@ -61,20 +61,45 @@ function AD.hessian(ba::AD.ForwardDiffBackend, f, x::AbstractArray)
...
     result = DiffResults.GradientResult(x)
     cfg = ForwardDiff.GradientConfig(f, x, chunk(ba, x))
     ForwardDiff.gradient!(result, f, x, cfg)
     return DiffResults.value(result), (DiffResults.derivative(result),)
 end
 
+function AD.value_and_second_derivative(ba::AD.ForwardDiffBackend, f, x::Real)
+    T = typeof(ForwardDiff.Tag(f, typeof(x)))
+    xdual = ForwardDiff.Dual{T}(x, one(x))
+    T2 = typeof(ForwardDiff.Tag(f, typeof(xdual)))
```
gdalle, line 80: can you explain how this works? tags are not documented anywhere

ext/AbstractDifferentiationForwardDiffExt.jl
```diff
@@ -61,20 +61,45 @@ function AD.hessian(ba::AD.ForwardDiffBackend, f, x::AbstractArray)
...
+    ydual = f(ForwardDiff.Dual{T2}(xdual, one(xdual)))
+    return ForwardDiff.value(T, ForwardDiff.value(T2, ydual)),
```
gdalle, line 82: perhaps intermediate variables would clarify things

gdalle:
As soon as I understand what's going on with the tags, it's good to merge on my end

ext/AbstractDifferentiationForwardDiffExt.jl
gerlero: TBH I'm cargo-culting this from `ForwardDiff`:

https://github.com/JuliaDiff/ForwardDiff.jl/blob/5e9240ae63e5d4217e1dc17cc14deca1d74c1ad7/src/derivative.jl#L12-L15

Tagging is there to [avoid perturbation confusion](https://juliadiff.org/ForwardDiff.jl/stable/user/advanced/#Custom-tags-and-tag-checking) in the presence of nested differentiation calls. Indeed there's no additional documentation and unfortunately that's pretty much the extent to which I'm familiar with this functionality.

In `ForwardDiff.derivative`, the `Tag` type is created from the function and the type of the input; so I'm copying exactly that behavior for higher-order differentiation too...

devmotion: But generally you may want other tags (e.g., SciML and Turing use custom tags), see https://www.stochasticlifestyle.com/improved-forwarddiff-jl-stacktraces-with-package-tags/

gdalle: I think @devmotion might be more qualified than us to approve of this part

ext/AbstractDifferentiationForwardDiffExt.jl
gerlero: I've added some variables to clarify the returns

test/test_utils.jl
```diff
@@ -6,6 +6,7 @@ Random.seed!(1234)
 fder(x, y) = exp(y) * x + y * log(x)
 dfderdx(x, y) = exp(y) + y * 1 / x
 dfderdy(x, y) = exp(y) * x + log(x)
+dfderdxdx(x, y) = -y * 1 / x^2
```
devmotion, line 9: Is there a reason to prefer this over the IMO more natural
```suggestion
dfderdxdx(x, y) = -y / x^2
```
?

test/test_utils.jl
```diff
@@ -143,6 +144,49 @@ function test_jacobians(backend; multiple_inputs=true, test_types=true)
...
+    else
+        # explicit test that AbstractDifferentiation throws an error
+        # don't support tuple of second derivatives
+        @test_throws ArgumentError AD.second_derivative(
+            backend, x -> fder(x, yscalar), (xscalar, yscalar)
+        )
+        @test_throws MethodError AD.second_derivative(
+            backend, x -> fder(x, yscalar), xscalar, yscalar
+        )
+    end
```
devmotion, line 160: Just remove the keyword argument if it is not supported?
```suggestion
function test_second_derivatives(backend; test_types=true)
    # explicit test that AbstractDifferentiation throws an error
    # don't support tuple of second derivatives
    @test_throws ArgumentError AD.second_derivative(
        backend, x -> fder(x, yscalar), (xscalar, yscalar)
    )
    @test_throws MethodError AD.second_derivative(
        backend, x -> fder(x, yscalar), xscalar, yscalar
    )
    end
```

test/test_utils.jl
```diff
@@ -143,6 +144,49 @@ function test_jacobians(backend; multiple_inputs=true, test_types=true)
...
+
+    # test if single input (no tuple works)
+    dder1 = AD.second_derivative(backend, x -> fder(x, yscalar), xscalar)
+    if test_types
+        @test dder1[1] isa Float64
```
devmotion, line 165: A bit stricter:
```suggestion
        @test only(dder1) isa Float64
```

test/test_utils.jl
```diff
@@ -143,6 +144,49 @@ function test_jacobians(backend; multiple_inputs=true, test_types=true)
...
+    end
+    @test dfderdxdx(xscalar, yscalar) ≈ dder1[1] atol = 1e-8
```
devmotion, line 167:
```suggestion
    @test dfderdxdx(xscalar, yscalar) ≈ only(dder1) atol = 1e-8
```

test/test_utils.jl
```diff
@@ -143,6 +144,49 @@ function test_jacobians(backend; multiple_inputs=true, test_types=true)
...
+    valscalar, dder2 = AD.value_and_second_derivative(
+        backend, x -> fder(x, yscalar), xscalar
+    )
+    if test_types
+        @test valscalar isa Float64
+        @test dder2[1] isa Float64
```
devmotion, line 173:
```suggestion
        @test only(dder2) isa Float64
```

test/test_utils.jl
```diff
@@ -143,6 +144,49 @@ function test_jacobians(backend; multiple_inputs=true, test_types=true)
...
+    end
+    @test valscalar == fder(xscalar, yscalar)
+    @test norm.(dder2 .- dder1) == (0,)
```
devmotion, line 176: Isn't `norm` unnecessarily complicated here?
```suggestion
    @test dder2 == dder1
```

test/test_utils.jl
```diff
@@ -143,6 +144,49 @@ function test_jacobians(backend; multiple_inputs=true, test_types=true)
...
+    valscalar, der, dder3 = AD.value_derivative_and_second_derivative(
+        backend, x -> fder(x, yscalar), xscalar
+    )
+    if test_types
+        @test valscalar isa Float64
+        @test der[1] isa Float64
+        @test dder3[1] isa Float64
```
devmotion, line 183:
```suggestion
        @test only(der) isa Float64
        @test only(dder3) isa Float64
```

test/test_utils.jl
```diff
@@ -143,6 +144,49 @@ function test_jacobians(backend; multiple_inputs=true, test_types=true)
...
+    end
+    @test valscalar == fder(xscalar, yscalar)
+    @test norm.(der .- AD.derivative(backend, x -> fder(x, yscalar), xscalar)) == (0,)
+    @test norm.(dder3 .- dder1) == (0,)
```
devmotion, line 187:
```suggestion
    @test der == AD.derivative(backend, x -> fder(x, yscalar), xscalar))
    @test dder3 == dder1
```

src/AbstractDifferentiation.jl
```diff
@@ -85,6 +85,24 @@ function jacobian(ab::HigherOrderBackend, f, xs...)
     return jacobian(lowest(ab), f, xs...)
 end
 
+"""
+    AD.second_derivative(ab::AD.AbstractBackend, f, xs...)
```
devmotion, line 89: The docstring only mentions `x` and the function only supports a single input:
```suggestion
    AD.second_derivative(ab::AD.AbstractBackend, f, x)
```

src/AbstractDifferentiation.jl
```diff
@@ -139,12 +157,23 @@ function value_and_jacobian(ab::AbstractBackend, f, xs...)
     return value, jacs
 end
 
+"""
+    AD.value_and_second_derivative(ab::AD.AbstractBackend, f, x)
+
+Return the tuple `(v, d2)` of the function value `v = f(x)` and the second derivatives `d = AD.derivative(ab, f, x)` and `d2 = AD.hessian(ab, f, x)`.
```
devmotion, line 163:
```suggestion
Return the tuple `(v, d2)` of the function value `v = f(x)` and the second derivative `d2 = AD.second_derivative(ab, f, x)`.
```

src/AbstractDifferentiation.jl
```diff
@@ -176,6 +205,28 @@ function value_and_hessian(ab::HigherOrderBackend, f, x)
     return value, hess
 end
 
+"""
+    AD.value_derivative_and_second_derivative(ab::AD.AbstractBackend, f, x)
+
+Return the tuple `(v, d, d2)` of the function value `v = f(x)` and the first and second derivatives `d = AD.derivative(ab, f, x)` and `d2 = AD.second_derivative(ab, f, x)`.
```
devmotion, line 211:
```suggestion
Return the tuple `(v, d, d2)` of the function value `v = f(x)`, the first derivative `d = AD.derivative(ab, f, x)`, and the second derivative `d2 = AD.second_derivative(ab, f, x)`.
```

ext/AbstractDifferentiationForwardDiffExt.jl
gdalle: @devmotion can we merge this and possibly discuss custom tags in a different PR?

src/AbstractDifferentiation.jl
gerlero: Good catch

test/test_utils.jl
gerlero: No reason; I was just following the style from above

devmotion (approved):

gerlero:
Thanks. Can we get a new release with this feature?

devmotion:
https://github.com/JuliaRegistries/General/pull/100271

Issue #124, opened 2023-12-21
`value_and_pullback_function` doesn't do what its docstring says


gdalle:
It returns `val, (tangents -> pullbacks)` instead of `tangents -> (val, pullbacks)`

For comparison, the forward equivalent does `tangents -> (val, pushforwards)` which is what it promises

ToucheSir:
I recall this came up before, and the idea was that this returns a tuple of "value" and "pullback function" rather than a "value and pullback"-returning function. Certainly the former seems to make more sense when we're talking about split reverse mode AD?

gdalle:
Regardless of what the best interpretation would be (I personally find this ambiguity horrifying), I think it would be good to fix the docstring first instead of changing the behavior. If you wanna review the associated PR #125 I can merge it afterwards :)

ToucheSir:
You'll find no complaints from me there, not sure why it wasn't called e.g. `value_and_pullback` in the first place!

Issue #119, opened 2023-10-18
Inconsistency in value_and_pushforward_function vs value_and_pullback_function outputs


prbzrg:
`value_and_pushforward_function` returns a function that returns `(value, pf)` 
https://github.com/JuliaDiff/AbstractDifferentiation.jl/blob/211b67528c5ed91971bb524c57adb63837163367/src/AbstractDifferentiation.jl#L259
but `value_and_pullback_function` returns `(value, pullback_function)`
https://github.com/JuliaDiff/AbstractDifferentiation.jl/blob/211b67528c5ed91971bb524c57adb63837163367/src/AbstractDifferentiation.jl#L319

devmotion:
I'm not sure there's anything to be fixed here. In forward mode usually you get both the primal and the dual together vs in reverse mode you get the value and the pullback function (see, e.g., ForwardDiff or ChainRulesCore.frule vs Zygote or ChainRulesCore.rrule).

prbzrg:
IMO, having a backend-agnostic interface means we shouldn't rely on backend APIs or mirror them. Moreover, every other value_and_AD APIs in `AbstractDifferentiation.jl` return value and AD.

prbzrg:
And why do we return the same value every time?

devmotion:
IMO the names are possibly confusing but not wrong per se - `value_and_X_function` can be read both as `(value_and_X)_function` and `value_and_(X_function)`. For forward mode we use the former interpretation and for reverse mode the latter.

AFAICT this difference between forward and reverse mode is not based on a specific package but a conceptual difference: In forward mode, you can (and usually do) compute both the primal value and the partial derivatives in a single forward path whereas in reverse mode you compute the primal value in the forward path, construct the pullback functions as you go, and then compute the derivatives in the backward path.

prbzrg:
Thanks for your response.

gdalle:
The docstring is misleading then, see #124 

That might be my fault

PR #125, opened 2023-12-21
Fix docstring for `value_and_pullback_function`

gdalle:
Fix the docstring of `value_and_pullback_function` to solve #124 and #119

Clarify that 
- `value_and_pushforward_function` means (value and pushforward) function
- `value_and_pullback_function` means value and (pullback function)

Name some anonymous functions in the process to make the code more readable

### DIFF
--- a/src/AbstractDifferentiation.jl
+++ b/src/AbstractDifferentiation.jl
@@ -226,44 +226,48 @@ The pushfoward function `pf` accepts as input a `Tuple` of tangents, one for eac
 If `xs` consists of a single element, `pf` can also accept a single tangent instead of a 1-tuple.
 """
 function pushforward_function(ab::AbstractBackend, f, xs...)
-    return (ds) -> begin
-        return jacobian(
-            lowest(ab),
-            (xds...,) -> begin
-                if ds isa Tuple
-                    @assert length(xs) == length(ds)
-                    newxs = xs .+ ds .* xds
-                    return f(newxs...)
-                else
-                    newx = only(xs) + ds * only(xds)
-                    return f(newx)
-                end
-            end,
-            _zero.(xs, ds)...,
-        )
+    function pf(ds)
+        function pf_aux(xds...)
+            if ds isa Tuple
+                @assert length(xs) == length(ds)
+                newxs = xs .+ ds .* xds
+                return f(newxs...)
+            else
+                newx = only(xs) + ds * only(xds)
+                return f(newx)
+            end
+        end,
+        return jacobian(lowest(ab), pf_aux, _zero.(xs, ds)...)
     end
+    return pf
 end
 
 """
     AD.value_and_pushforward_function(ab::AD.AbstractBackend, f, xs...)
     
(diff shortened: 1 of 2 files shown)


src/AbstractDifferentiation.jl
```diff
@@ -226,44 +226,48 @@ The pushfoward function `pf` accepts as input a `Tuple` of tangents, one for eac
...
-                    newxs = xs .+ ds .* xds
-                    return f(newxs...)
-                else
-                    newx = only(xs) + ds * only(xds)
-                    return f(newx)
-                end
-            end,
-            _zero.(xs, ds)...,
-        )
+    function pf(ds)
```
devmotion, line 229: Generally I would have preferred to not include such changes in the PR as they are not related to the docstring fixes and make the PR more difficult to review.

src/AbstractDifferentiation.jl
```diff
@@ -291,21 +295,24 @@ The pullback function `pb` accepts as input a `Tuple` of cotangents, one for eac
 If `f` has a single output, `pb` can also accept a single input instead of a 1-tuple.
 """
 function pullback_function(ab::AbstractBackend, f, xs...)
-    _, pbf = value_and_pullback_function(ab, f, xs...)
-    return pbf
+    _, pb = value_and_pullback_function(ab, f, xs...)
```
devmotion, line 298: What's the motivation for this change? `pbf` seemed to refer to "pullback function" I assume which seems to be the correct interpretation?

src/AbstractDifferentiation.jl
gdalle: Generally I would agree but I included this refactoring because it makes this part of the code clearer. I was the one who wrote all the docstrings the first time around, and I screwed up on `value_and_pullback_function` partly because the code was a bit obscure

src/AbstractDifferentiation.jl
gdalle: I think the terms "pullback" and "pullback function" are both valid, see eg. the JAX docs. My rationale for this was to fight ambiguities, because the function returning pb is already called pullback_function. So the idea is that the pullback function returns a pullback, which is itself a function

src/AbstractDifferentiation.jl
devmotion: Yeah, I'm fine with it being included in this PR but wanted to mention it nevertheless as a general consideration.

src/AbstractDifferentiation.jl
devmotion: My feeling was that the change here actually made it less clear - `gradient` etc. return gradients, and similarly `pullback_function` returns a "pullback function" for which `pbf` seemed to be an appropriate abbreviation.

src/AbstractDifferentiation.jl
gdalle: Then we're in agreement :)

src/AbstractDifferentiation.jl
gdalle: Then we should also turn `pf` into `pff` on the pushforward side (another reason for my change). I don't have strong opinions either way, it just feels weird to me that `pullback_function` returns a `pbf` which is itself short for... pullback function

gdalle: fixed in my latest commit

devmotion (approved):

PR #128, opened 2024-02-04
Use `Zygote.jacobian` etc.

### LINKED ISSUE #54: `AD.jacobian` much slower than `Zygote.jacobian`
Hi, and thanks for this amazing interface!

When computing jacobians, I recently noted a significant speed difference  between standalone Zygote and Zygote used as an AD backend. The allocations also differ wildly. Do you happen to know where that comes from?

Here's a minimal working example:

```julia
julia> using AbstractDifferentiation, BenchmarkTools, Zygote

julia> f(x) = x .^ 2
f (generic function with 1 method)

julia> x = rand(100);

julia> ab = AD.ZygoteBackend()
AbstractDifferentiation.ReverseRuleConfigBackend{Zygote.ZygoteRuleConfig{Zygote.Context}}(Zygote.ZygoteRuleConfig{Zygote.Context}(Zygote.Context(nothing)))

julia> Zygote.jacobian(f, x)
([1.3044988022198039 0.0 … 0.0 0.0; 0.0 1.9710054432976252 … 0.0 0.0; … ; 0.0 0.0 … 1.9276513864536091 0.0; 0.0 0.0 … 0.0 1.6930015760093526],)

julia> AD.jacobian(ab, f, x)
([1.3044988022198039 0.0 … 0.0 0.0; 0.0 1.9710054432976252 … 0.0 0.0; … ; 0.0 0.0 … 1.9276513864536091 0.0; 0.0 0.0 … 0.0 1.6930015760093526],)

julia> @benchmark Zygote.jacobian(f, x)
BenchmarkTools.Trial: 7001 samples with 1 evaluation.
 Range (min … max):  629.524 μs …   5.403 ms  ┊ GC (min … max): 0.00% … 87.03%
 Time  (median):     681.270 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   712.290 μs ± 295.183 μs  ┊ GC (mean ± σ):  3.42% ±  6.90%

                    ▃█▄                                          
  ▁▁▁▁▁▁▂▁▁▁▁▂▂▂▂▁▁▃███▇▄▄▃▅▅▅▄▄▃▃▃▃▂▂▂▂▂▂▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  630 μs           Histogram: frequency by time          778 μs <

 Memory estimate: 427.06 KiB, allocs estimate: 4141.

julia> @benchmark AD.jacobian(ab, f, x)
BenchmarkTools.Trial: 1393 samples with 1 evaluation.
 Range (min … max):  2.673 ms … 64.564 ms  ┊ GC (min … max):  0.00% … 94.96%
 Time  (median):     2.999 ms              ┊ GC (median):     0.00%
 Time  (mean ± σ):   3.583 ms ±  4.786 ms  ┊ GC (mean ± σ):  11.42% ±  8.30%

  ▂▇██▆▆▅▄▃▂                                                  
  █████████████▆▇█▆▅▅▅▅▅▅▅▅▄▅▁▆▁▁▁▁▄▄▁▄▁▄▁▅▄▄▁▁▁▄▁▅▄▄▁▁▁▁▄▄▄ █
  2.67 ms      Histogram: log(frequency) by time      8.1 ms <

 Memory estimate: 1.26 MiB, allocs estimate: 28023.
```

devmotion:
Zygote already has an interface for pullbacks, jacobians, etc., hence IMO we should use it in `AD.jacobian` etc. as I guess users (arguably correctly) expect that `AD.jacobian` performs the same computations and is as efficient as `Zygote.jacobian`. That is, I think AbstractDifferentiation should not treat Zygote as "just" an example of ChainRules AD backend but rather as a separate entity as - I think - we should not redefine jacobian etc. for Zygote based on the ChainRules pullback with `Zygote.ZygoteRuleConfig`.

Fixes #54.

### DIFF
--- a/ext/AbstractDifferentiationZygoteExt.jl
+++ b/ext/AbstractDifferentiationZygoteExt.jl
@@ -8,11 +8,27 @@ else
     using ..Zygote: Zygote
 end
 
-AD.ZygoteBackend() = AD.ReverseRuleConfigBackend(Zygote.ZygoteRuleConfig())
-
 # Context should not persist between different AD calls: fixes #69
 function AD.ruleconfig(::AD.ReverseRuleConfigBackend{<:Zygote.ZygoteRuleConfig})
     return Zygote.ZygoteRuleConfig()
 end
+
+function AD.value_and_pullback_function(::AD.ZygoteBackend, f, args...)
+    return Zygote.pullback(f, args...)
+end
+
+AD.gradient(::AD.ZygoteBackend, f, args...) = Zygote.gradient(f, args...)
+function AD.value_and_gradient(::AD.ZygoteBackend, f, args...)
+    res = Zygote.withgradient(f, args...)
+    return res.val, res.grad
+end
+
+AD.jacobian(::AD.ZygoteBackend, f, args...) = Zygote.jacobian(f, args...)
+function AD.value_and_jacobian(::AD.ZygoteBackend, f, args...)
+    res = Zygote.withjacobian(f, args...)
+    return res.val, res.grad
+end
+
+AD.hessian(::AD.ZygoteBackend, f, arg) = Zygote.hessian(f, arg)
 
 end # module
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -71,13 +71,14 @@ end
 ruleconfig(ba::ReverseRuleConfigBackend) = ba.ruleconfig
 
 """
-    ZygoteBackend()
+    ZygoteBackend
 
 Create an AD backend that uses reverse mode with [Zygote.jl](https://github.com/FluxML/Zygote.jl).
 
-It is a special case of [`ReverseRuleConfigBackend`](@ref).
+Alternatively, you can perform AD with Zygote using a special [`ReverseRuleConfigBackend`](@ref), namely `ReverseRuleConfigBackend(Zygote.ZygoteRuleConfig())`.
+Note, however, that the behaviour of this backend is not equivalent to `ZygoteBackend()` since the former uses a generic implementation of jacobian etc. for ChainRules-compatible AD backends whereas `ZygoteBackend` uses implementations in Zygote.jl.
 
 !!! note
     To be able to use this backend, you have to load Zygote.
 """
-function ZygoteBackend end
+struct ZygoteBackend <: AbstractReverseMode end
--- a/test/ruleconfig.jl
+++ b/test/ruleconfig.jl
@@ -4,7 +4,10 @@ using Test
 using Zygote
 
 @testset "ReverseRuleConfigBackend(ZygoteRuleConfig())" begin
-    backends = [@inferred(AD.ZygoteBackend())]
+    backends = [
+        @inferred(AD.ZygoteBackend()),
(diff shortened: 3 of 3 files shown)


mohamed82008:
IIRC, one fundamental difference is that Zygote.jacobian preallocates the Jacobian matrix. This is not good for higher order derivatives with reverse mode because it is mutating.

devmotion:
My feeling is that this is a problem that should be addressed in Zygote (possibly they should use the default implementation in AbstractDifferentiation based on `value_and_pullback_function` but that's their decision). Based on the number of open (and closed) issues about AbstractDifferentiation taking different code paths, and how it leads to sometimes different behaviour and sometimes different (typically worse) performance, I think if a function such as `jacobian`, `hessian` etc. exists in an AD backend it should be used by AbstractDifferentiation.

gdalle:
I agree with this take, and I push it further here: 

https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/129

mohamed82008:
`Zygote.hessian` is arguably even more questionable because it actually uses ForwardDiff for the second derivative.

devmotion:
My view is the same there as well: If `Zygote.hessian` is broken/bad, an issue should be opened in Zygote and possibly they might want to switch to hessian definition for pullbacks in AbstractDifferentiation. But we should not fix AD backends in AbstractDifferentiation.

devmotion:
Bump 🙂 

For other AD backends such as ForwardDiff, ReverseDiff, and Tracker actually we already use jacobian, hessian, etc. in downstream packages when applicable.

ext/AbstractDifferentiationZygoteExt.jl
```diff
@@ -8,11 +8,27 @@ else
...
+    return Zygote.pullback(f, args...)
+end
+
+AD.gradient(::AD.ZygoteBackend, f, args...) = Zygote.gradient(f, args...)
+function AD.value_and_gradient(::AD.ZygoteBackend, f, args...)
+    res = Zygote.withgradient(f, args...)
+    return res.val, res.grad
+end
+
+AD.jacobian(::AD.ZygoteBackend, f, args...) = Zygote.jacobian(f, args...)
```
mohamed82008, line 26: can we provide a way to opt out of this, perhaps a type parameter?

devmotion: We don't provide any opt-outs for any other backend, so my feeling is that we should not add one here. I think it would also unnecessarily complicate the backend (and if there's a problem with `Zygote.jacobian` I still think it should be fixed in Zygote). If a user really wants to use a different implementation of `AD.jacobian` I think it should be reasonably easy to define a custom `MyZygoteBackend` that forwards everything but `AD.jacobian` to a `ZygoteBackend`.

mohamed82008: I might be the minority vote here in which case I will take the extra inconvenience of defining my own backend.

gdalle: I agree with David on this one

mohdibntarek (approved):

